PR #6, opened 2019-09-10
Fix uninitialized variable in FvwmIconMan colorset loop.

cyberbisson:
In FvwmIconMan, using colorsets to set various styles (e.g., PlainColorset,
SelectColorset, etc.) is broken.  Essentially, background colors for one color
end up being mapped to the shadow color of the previously used colorset.  This
is because the loop that calls fvwmlib_XCreateGC on the various colorsets does
not happen to set the background color from the colorset inside the loop.  This
makes the code function OK with just one colorset, but adding a second colorset
in your settings causes problems.

This change explicitly set the background color on the GC within the loop, so as
to not carry forth the shadow color from the end of the previous iteration.


ThomasAdam:
Thank you!

PR #10, opened 2020-01-06
Fix converting of multi-images icon files

### LINKED ISSUE #9: fvwm-menu-desktop is wrong when converting multiple image files.
When an original icon contain several images, convert in fvwm-menu-desktop is converting all the images and rename them into "filename-n.png", where n is the number of the image into the original file.

The result is a broken resulting menu because fvwm look for filename.png.

To reproduce it, I find 2 icons in my system, the ones from doxbox and xmahjongg. Otherwise any animated gif will do it.

It must be the same issue with fvwm2 as well. That patch fix that issue:

[fix_multiple_images_file_conversion.patch.txt](https://github.com/fvwmorg/fvwm3/files/4027355/fix_multiple_images_file_conversion.patch.txt)

domichel:
Fix #9

### DIFF
--- a/bin/fvwm-menu-desktop.in
+++ b/bin/fvwm-menu-desktop.in
@@ -2,6 +2,10 @@
  
 # Modification History
 
+# Changed on 06/01/20 by Dominique Michel:
+# - Fix converting of multi-images icon files;
+#   use the first one.
+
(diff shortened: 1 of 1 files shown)


domichel:
Just close it.

PR #36, opened 2020-03-19
should not compare menu positions with screen width/height.

mikeandmore:
Under a vertical dual display setup, menu would jump from the lower display to the higher one when invoking `MenuMoveCursor`. This patch fixes that.

In general, we should not compare a position with the screen width/height. Screens have x/y positions too.


ThomasAdam:
Thanks -- this wasn't a problem before, because of Xinerama.

Issue #42, opened 2020-04-11
Do we need to check the value returned from FCreateFImage?


klebertarcisio:
Hi everyone,

I would like to know if we need any verification after calling the FCreateFImage function. See the lines below that it can return a value that has already been released.

https://github.com/fvwmorg/fvwm3/blob/6213e5ebc76c43da667b4b13012c09aa83a58bb7/libs/FImage.c#L188-L193

See that in the code below there is a check on the value of out_fim

https://github.com/fvwmorg/fvwm3/blob/6213e5ebc76c43da667b4b13012c09aa83a58bb7/libs/PictureGraphics.c#L851-L874

Why doesn't the same check exist on the line below?

https://github.com/fvwmorg/fvwm3/blob/6213e5ebc76c43da667b4b13012c09aa83a58bb7/libs/PictureGraphics.c#L577-L578

ThomasAdam:
Hi,

Thank you.  Please open a pull-request against this issue, and I'll take a look at any code change you care to make.

Kindly,
Thomas

ThomasAdam:
Kindly,
Thomas

PR #43, opened 2020-04-11
Do we need to check the value returned from FCreateFImage?

klebertarcisio:
This pull-request is about this issue #42


ThomasAdam:
Looks good.

PR #90, opened 2020-05-23
remove dependency on Perl4::CoreLibs::getcwd (fvwm3)

slazav:
Cwd module is used instead of outdated getcwd.pl from Perl4::CoreLibs
see: https://metacpan.org/pod/Perl4::CoreLibs
To keep old behaviour "\n" is added to the output.

### DIFF
--- a/perllib/General/FileSystem.pm
+++ b/perllib/General/FileSystem.pm
@@ -1066,7 +1066,7 @@ Returns the current working directory.
 
 
 sub get_cwd () {
-	$^O eq "MSWin32" ? Win32::GetCwd() : require "getcwd.pl" && getcwd();
+	$^O eq "MSWin32" ? Win32::GetCwd() : require Cwd && getcwd()."\n";
 }
 
 


ThomasAdam:
Hi @slazav,

Thanks for this. It would be good if you could take a look at the minimum perl version we assert in `configure.ac` (5.8), and take steps to bump the minimum version we support to 5.20.x

Can you augment this patch series to look into this?

Thanks,
Thomas

slazav:
> It would be good if you could take a look at the minimum perl version we assert in configure.ac 
> (5.8), and take steps to bump the minimum version we support to 5.20.x

I'm not familiar with perl versions and do not quite understand what can be done here. In the configure.ac REQUIRED_PERL_VERSION=5.004, and Cwd::getcwd function exists in this version:
https://metacpan.org/pod/distribution/Module-CoreList-List/lib/Module/CoreList/List/5_004.pod
It looks like there is no need to increase it.

ThomasAdam:
Hi @slazav,

Well, there's things we can do...

- Deprecate `use fields`
- Remove subroutine prototypes
- Use `IPC::Run` over backticks

But your patch is such a small step towards this. I'll apply it, but should you want to work on the above, let me know.

Kindly,
Thomas

PR #113, opened 2020-06-02
key binding and menu option for dmenu_run

lgsobalvarro:
Adds Meta+Space as trigger for dmenu_run
Tests the existence of dmenu and if present Adds "Run command" into fvwm menu. (Probably a run command icon is required) to accompany the menu.


ThomasAdam:
Thanks!

PR #126, opened 2020-06-07
Add FVWM3_LOGDIR envvar for logging to different file

NsCDE:
Hi Thomas,

I don't know what was more difficult: to make my 1st PR in life or how to do ~/ substitution.

I have put define in log.h and renamed env variable to FVWM3_LOGDIR as you requested. Please review. I hope this is ok.

### DIFF
--- a/libs/log.c
+++ b/libs/log.c
@@ -30,6 +30,10 @@
 #include "fvwmlib.h"
 #include "log.h"
 
+#ifdef HAVE_GETPWUID
+#include <pwd.h>
+#endif
+
 static FILE	*log_file;
 static int	 log_level;
 
@@ -46,7 +50,31 @@ log_set_level(int ll)
 void
 log_open(void)
 {
-	char	*path = "fvwm3-output.log";
+	char	*path;
+
+	if (getenv("FVWM3_LOGFILE"))
+	{
+		path = fxstrdup(getenv("FVWM3_LOGFILE"));
+#ifdef HAVE_GETPWUID
+		if (strncmp(path, "~/", 2) == 0)
+		{
+		        /* Remove only ~ we need / between homedir and the rest */
+			memmove(path, path + 1, strlen(path));
+
+			struct passwd *pwd;
+
+	                pwd = getpwuid(getuid());
+			if (pwd)
+			{
+				path = strcat(pwd->pw_dir, path);
+			}
+		}
+#endif
+	}
+	else
+	{
+		path = FVWM3_LOGFILE_DEFAULT;
+	}
 
 	if (log_level == 0)
 		return;
--- a/libs/log.h
+++ b/libs/log.h
@@ -2,6 +2,7 @@
 #define _LOG_H
 
 #define printflike(a, b) __attribute__ ((format (printf, a, b)))
+#define FVWM3_LOGFILE_DEFAULT "fvwm3-output.log"
 
 void	log_set_level(int);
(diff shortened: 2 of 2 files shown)


libs/log.c
```diff
@@ -46,7 +50,31 @@ log_set_level(int ll)
...
+	char	*path;
+
+	if (getenv("FVWM3_LOGFILE"))
+	{
+		path = fxstrdup(getenv("FVWM3_LOGFILE"));
+#ifdef HAVE_GETPWUID
+		if (strncmp(path, "~/", 2) == 0)
+		{
+		        /* Remove only ~ we need / between homedir and the rest */
+			memmove(path, path + 1, strlen(path));
```
ThomasAdam, line 62: Excellent start.  This `memmove` though will overrun if path isn't NUL-terminated.  `strlen(...) + 1` will fix that.

libs/log.c
```diff
@@ -46,7 +50,31 @@ log_set_level(int ll)
...
+
+			struct passwd *pwd;
+
+	                pwd = getpwuid(getuid());
+			if (pwd)
+			{
+				path = strcat(pwd->pw_dir, path);
```
ThomasAdam, line 69: `path` is allocated via `fxstrdup()` and needs freeing here, if it's to be allocated in this way, via  temporary variable, otherwise we'll leak memory.

NsCDE: Hi Thomas,

I'm glad you expanded this and got idea for consolidating generally things about homedir, shell paths and repetative usage of getpwuid in fvwm code.

Yesterday, while searching for similart things in the code, I saw fvwm/session.c, libs/fsm.c, and modules/FvwmScript/Instructions.c are calling getpwuid to find pw->pw_dir too. Since you produced the whole centralized .c+.h files for handling canonification of homedir and path in all the code, this can be consolidated too.

For a C language, I am what people here is Balkan like to say "I'm not a gynecologist, but I can take a look". :-) That said, I will be more carefull to free() things in the future. I have one [patch](https://github.com/NsCDE/NsCDE/blob/master/src/FvwmButtons_sunkraise_windowname_unified.patch) for FvwmButtons (window name and indicator) for a long time too, but this can probably wait post 1.0, it needs to be revisited for fvwm3 ...

P. S.
Simply drop this and apply #127 

Cheers ...

ThomasAdam: > Yesterday, while searching for similart things in the code, I saw fvwm/session.c, libs/fsm.c, and modules/FvwmScript/Instructions.c are calling getpwuid to find pw->pw_dir too. Since you produced the whole centralized .c+.h files for handling canonification of homedir and path in all the code, this can be consolidated too.

Well, feel free to transition those callers across to this, too -- no rush though.
 
> For a C language, I am what people here is Balkan like to say "I'm not a gynecologist, but I can take a look". :-) 

Heh -- an interesting sentiment there.  :)

> That said, I will be more carefull to free() things in the future. I have one [patch](https://github.com/NsCDE/NsCDE/blob/master/src/FvwmButtons_sunkraise_windowname_unified.patch) for FvwmButtons (window name and indicator) for a long time too, but this can probably wait post 1.0, it needs to be revisited for fvwm3 ...

Do please open up a pull-request with that patch, against `FVWM3` and I'll take a look...

> P. S.
> Simply drop this and apply #127

Will do.  Thanks again.

ThomasAdam:
Closing per review comments.  Superseded by #127

PR #137, opened 2020-06-24
Update to config

lgsobalvarro:
- IconMan now uses colorset 10 instead 0, to honor Wiki conventions (Thanks to somiaj)
- dmenu_run uses -nb '#2b4e5e' so it'll integrate better with the new colorset (Suggestion by Master_P_the_Gu)
- Edited menu option xemacs for emacs since xemacs is in maintenance mode and emacs seems to be more popular this days.

### DIFF
--- a/default-config/config
+++ b/default-config/config
@@ -285,7 +285,7 @@ AddToMenu   MenuFvwmRoot "Fvwm" Title
 + "&Programs%icons/programs.png%" Popup MenuPrograms
 + "XDG &Menu%icons/apps.png%" Popup XDGMenu
 + "&XTerm%icons/terminal.png%" Exec exec $[infostore.terminal]
-Test (x dmenu_run) + "Run command" Exec exec dmenu_run
+Test (x dmenu_run) + "Run command" Exec exec dmenu_run -nb "#2b4e5e"
 + "" Nop
 + "Fvwm&Console%icons/terminal.png%" Module FvwmConsole -terminal $[infostore.terminal]
 + "&Wallpapers%icons/wallpaper.png%" Popup BGMenu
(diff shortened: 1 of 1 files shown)


default-config/config
```diff
@@ -285,7 +285,7 @@ AddToMenu   MenuFvwmRoot "Fvwm" Title
 + "&Programs%icons/programs.png%" Popup MenuPrograms
 + "XDG &Menu%icons/apps.png%" Popup XDGMenu
 + "&XTerm%icons/terminal.png%" Exec exec $[infostore.terminal]
-Test (x dmenu_run) + "Run command" Exec exec dmenu_run
+Test (x dmenu_run) + "Run command" Exec exec dmenu_run -nb "#2b4e5e"
```
ThomasAdam, line 288: We should move this colour definition to using infostore, and use it here.

default-config/config
```diff
@@ -462,7 +462,7 @@ Silent Key F2 A C GotoDesk 0 1
 Silent Key F3 A C GotoDesk 0 2
 Silent Key F4 A C GotoDesk 0 3
 Silent Key Super_R A A Exec exec $[infostore.terminal]
-Silent Key Space A M Exec exec dmenu_run
+Silent Key Space A M Exec exec dmenu_run -nb "#2b4e5e"
```
ThomasAdam, line 465: Same here.

default-config/config
lgsobalvarro: That's a good suggestion, it would simplify things. Do you suggest only changing the color definition or use infostore for the full command? Perhaps the full command option could be more convenient for those who want to switch to Rofi or something.

ThomasAdam: Sure. Even better. Go for it!

PR #138, opened 2020-06-25
removed BroadcastMonitorList from HandleFocusIn:

d-e-e-p:
In HandleFocusIn do_force_broadcast gets set which forces BroadcastMonitorList to be sent to all modules every time focus is moved to a window.  As a result FvwmBacker reloads background screen way too often, even when monitor has not changed.  Perhaps monitor updates should be handled by monitor_update_ewmh ...

### DIFF
--- a/fvwm/events.c
+++ b/fvwm/events.c
@@ -2596,7 +2596,6 @@ void HandleFocusIn(const evh_args_t *ea)
 		last_focus_fw = focus_fw;
 		was_nothing_ever_focused = False;
 
-		BroadcastMonitorList(NULL);
 	}
 	if ((sf = get_focus_window()) != ffw_old)
 	{


ThomasAdam:
Hi @d-e-e-p,

Thank you for this PR, and for highlighting the breakage I've introduced to `FvwmBacker`.  Whoops, not intended!

Although I understand your suggestion, just blindly removing `BroadcastMonitorList()` from handling `FocusIn` events is not the correct solution since you still need to inform modules that a change of monitor focus has potentially occured, which is why this check is here.

However, the underlying bug which you've found is that consumers of this event (which is all modules right now) have no way of ignoring these.  Since we only care about a change in focused monitor, I think it best we:

* Introduce a `MX_MONITOR_FOCUS` flag to our extended module flags, so modules can express an interest in receiving this information (currently just `FvwmPager` but I might expand this to `FvwmEvent` at some point;
* Track the previously focused monitor and only send this event if the newly-focused monitor is not the same as the old one.

To this end, I've created this:  https://github.com/fvwmorg/fvwm3/pull/140

Please can you try this for me.  It's all on the `ta/mx_focus` branch if that's easier to test.

I'll therefore close this PR, and you can comment on #140 instead.

Thanks again!
Thomas

PR #139, opened 2020-06-25
Colorset variable expansion features.

somiaj:
```
  * Added appending .lighten<p> and .darken<p> to a colorset
    variable expansion, such as $[fg.cs<n>.lighten<p>]
    or $[shadow.cs<n>.darken<p>], that will lighten or darken
    the color by percentage p before outputing the color string.
  * Added appending .hash to the end of the colorset variable
    expansion to output the color as a hash, "#rrggbb", instead
    of the color form "rgb:rrrr/gggg/bbbb". Examples:
    $[bg.cs10.hash] or $[fgsh.cs5.darken20.hash].
  * Added missing $[fgsh.cs<n>] description to the manpage.
```

### DIFF
--- a/doc/fvwm/expansion.xml
+++ b/doc/fvwm/expansion.xml
@@ -441,15 +441,32 @@ command can be used:</para>
       $[bg.cs&lt;n&gt;]
       $[hilight.cs&lt;n&gt;]
       $[shadow.cs&lt;n&gt;]
+      $[fgsh.cs&lt;n&gt;]
     </term>
     <listitem>
       <para>
 	These parameters are replaced with the name of the foreground
-	(fg), background (bg), hilight (hilight) or shadow (shadow) color
-	that is defined in colorset &lt;n&gt; (replace &lt;n&gt; with zero
-	or a positive integer).  For example "$[fg.cs3]" is expanded to the
-	name of the foreground color of colorset 3 (in rgb:rrrr/gggg/bbbb
-	form).  Please refer to the
+	(fg), background (bg), hilight (hilight), shadow (shadow), or
+        the font shadow (fgsh) color that is defined in colorset &lt;n&gt;
+        (replace &lt;n&gt; with zero or a positive integer).  For example
+        "$[fg.cs3]" is expanded to the name of the foreground color of
+        colorset 3 (in rgb:rrrr/gggg/bbbb form).
+      </para>
+      <para>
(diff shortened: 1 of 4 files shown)


doc/fvwm/expansion.xml
```diff
@@ -441,15 +441,32 @@ command can be used:</para>
...
-	name of the foreground color of colorset 3 (in rgb:rrrr/gggg/bbbb
-	form).  Please refer to the
+	(fg), background (bg), hilight (hilight), shadow (shadow), or
+        the font shadow (fgsh) color that is defined in colorset &lt;n&gt;
+        (replace &lt;n&gt; with zero or a positive integer).  For example
+        "$[fg.cs3]" is expanded to the name of the foreground color of
+        colorset 3 (in rgb:rrrr/gggg/bbbb form).
+      </para>
+      <para>
+        If .lighten&lt;p&gt; or .darken&lt;p&gt; is appeneded to the
```
ThomasAdam, line 456:
```suggestion
        If .lighten&lt;p&gt; or .darken&lt;p&gt; is appended to the
```

fvwm/expand.c
```diff
@@ -391,6 +394,34 @@ static signed int expand_vars_extended(
...
+			if (sscanf(rest + n, ".lighten%d%n", &l, &nn) && l >= 0 && l <= 100)
+			{
+				csadj = l;
+			}
+			if (sscanf(rest + n, ".darken%d%n", &l, &nn) && l >= 0 && l <= 100)
+			{
+				csadj = -l;
+			}
+			/* Check for .hash */
+			if (strcmp(rest + n + nn,  ".hash") == 0)
```
ThomasAdam, line 410: I would suggest the following:

```suggestion
			if (StrEquals(rest + n + nn,  ".hash"))
```

`StrEquals()` doesn't care about case, so I can use: `.HaSh` if I wanted, but I can't with plain `strcmp()`.  I know I've been guilty of this myself in the past.  Need to change this.

You could also offset the string comparison to elide the `.`, i.e.:  `rest + n + nn + 1`, but it's probably not worth it.

fvwm/expand.c
```diff
@@ -391,6 +394,34 @@ static signed int expand_vars_extended(
...
+			{
+				use_hash = True;
+				nn = nn + 5;
```
ThomasAdam, line 413: Try and avoid any magic variables:

```suggestion
				nn += strlen(".hash");
```

PR #140, opened 2020-06-25
monitor: track focus event separately

ThomasAdam:
When modules learned of the ability to send monitor information, this
was previously sent on all window focus events, as the FVWM modules
required this to reset which monitor had the focus.

This is not ideal with other modules which end up consuming this
information when they might not care about monitor information at all.

Instead, remove BroadcastMonitor() from FocusIn events and set FvwmPager
to be a listener of MX_MONITOR_FOCUS.

### DIFF
--- a/fvwm/events.c
+++ b/fvwm/events.c
@@ -2444,6 +2444,7 @@ void HandleFocusIn(const evh_args_t *ea)
 	static Window last_focus_fw = None;
 	static Bool was_nothing_ever_focused = True;
 	FvwmWindow *fw = ea->exc->w.fw;
+	struct monitor *mon = monitor_get_current();
 
 	Scr.focus_in_pending_window = NULL;
 	/* This is a hack to make the PointerKey command work */
@@ -2591,12 +2592,17 @@ void HandleFocusIn(const evh_args_t *ea)
 				(unsigned long)IsLastFocusSetByMouse(),
 				(long)fc, (long)bc);
 			EWMH_SetActiveWindow(focus_w);
+
+			if (fw != NULL && strcmp(fw->m->si->name, prev_focused_monitor) != 0) {
+				BroadcastName(MX_MONITOR_FOCUS, -1, -1, -1,
+				    fw->m->si->name /* Name of the monitor. */
+			        );
+			}
 		}
 		last_focus_w = focus_w;
 		last_focus_fw = focus_fw;
 		was_nothing_ever_focused = False;
 
-		BroadcastMonitorList(NULL);
 	}
 	if ((sf = get_focus_window()) != ffw_old)
 	{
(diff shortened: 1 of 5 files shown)


d-e-e-p:
solves the FocusIn problem!  
modules still need a way to stop receiving MX* messages, eg FvwmBacker gets all MX messages unless we do something like:

```
 **modules/FvwmBacker/FvwmBacker.c**

 SetMessageMask(fvwm_fd,
              M_NEW_PAGE | M_NEW_DESK | M_CONFIG_INFO |
              M_END_CONFIG_INFO | M_SENDCONFIG );
 #define MX_NONE ((1<<9) | M_EXTENDED_MSG)
 SetMessageMask(fvwm_fd, MX_NONE);                                                                                   
  
```

ThomasAdam:
Hi @d-e-e-p,

Yes -- it seems as though modules by design are sent MX_* messages, despite the bitmask saying otherwise.

I am not going to change this as it could break things.

Equally, I'm not going to embed the bit offset information about packet ordering for MX_* messages in to `FvwmBacker`.

Is this having a real impact for FvwmBacker, since it's always been this way.

d-e-e-p:
I didn't measure the performance impact but it does seem like a waste to send all these MX messages that just get ignored downstream.

**Current:**

```
 FvwmAnimate    all
 FvwmAuto       MX_ENTER_WINDOW | MX_LEAVE_WINDOW
 FvwmBacker     all
 FvwmButtons    MX_PROPERTY_CHANGE
 FvwmConsole    all
 FvwmEvent      all
 FvwmIconMan    MX_VISIBLE_ICON_NAME | MX_PROPERTY_CHANGE
 FvwmIdent      MX_PROPERTY_CHANGE
 FvwmPager      MX_VISIBLE_ICON_NAME| MX_PROPERTY_CHANGE| MX_REPLY
 FvwmPerl       all
 FvwmRearrange  all
 FvwmScript     MX_PROPERTY_CHANGE
```


**Proposed:**

define a new MX_NONE in **libs/Module.h** and have an explicit SetMessageMask of MX_NONE in FvwmAnimate, FvwmBacker, FvwmConsole, FvwmRearrange :

```
 FvwmAnimate    MX_NONE
 FvwmAuto       MX_ENTER_WINDOW | MX_LEAVE_WINDOW
 FvwmBacker     MX_NONE
 FvwmButtons    MX_PROPERTY_CHANGE
 FvwmConsole    MX_NONE
 FvwmEvent      all (unchanged)
 FvwmIconMan    MX_VISIBLE_ICON_NAME | MX_PROPERTY_CHANGE
 FvwmIdent      MX_PROPERTY_CHANGE
 FvwmPager      MX_VISIBLE_ICON_NAME| MX_PROPERTY_CHANGE| MX_REPLY
 FvwmPerl       all (unchanged)
 FvwmRearrange  MX_NONE
 FvwmScript     MX_PROPERTY_CHANGE
```

so 4 modules would change.

ThomasAdam:
Hi @d-e-e-p,

It does, but it's not something I'm committing to doing until after the first release of fvwm3.

What I'm asking is if the original problem is fixed, and I'll mop up stuff like this afterwards.

d-e-e-p:
original problem is fixed!  thanks.

PR #155, opened 2020-07-01
Avoiding null reference

klebertarcisio:
Calloc() can returns NULL if there is not enough storage.

### DIFF
--- a/libs/FScreen.c
+++ b/libs/FScreen.c
@@ -348,13 +348,15 @@ scan_screens(Display *dpy)
 			m = monitor_new();
 			m->flags |= MONITOR_NEW;
 			m->si = screen_info_new();
-			m->si->name = strdup(name);
-			memset(&m->virtual_scr, 0, sizeof(m->virtual_scr));
+			if(m->si != NULL){
+				m->si->name = strdup(name);
+				memset(&m->virtual_scr, 0, sizeof(m->virtual_scr));
 
-			TAILQ_INSERT_TAIL(&screen_info_q, m->si, entry);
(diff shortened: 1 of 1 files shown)


ThomasAdam:
Hi @klebertarcisio,

Ah, that's a shame!  I've just pushed a fix for this problem (I've had it kicking around on a branch and just got round to looking at it).  Note that your change is unnecessary in that there's already a `fxcalloc()` wrapper which is the only change required here.

If you spot any other calls not using the `fx*()` wrappers I've put in place, do send through additional patches.

Cheers!
Thomas

klebertarcisio:
Hi @ThomasAdam, I'm talking about the **screen_info_new** function that returns an object without checking if it is null. Should we change this function?

ThomasAdam:
Hi @klebertarcisio,

Yes, I know what you're reffering to.  Please check the HEAD commit on `master`.  I've already changed it to use `fxcalloc()` which you'll find defined in `libs/safemalloc.c`.  It was an oversight on my part not to have used it to begin with.  You'll also note that `fxcalloc()` has many safeguards to ensure that the size and return types are correct, and more importantly, never NULL.

ThomasAdam:
If you wanted, @klebertarcisio -- you'll probably find a few places in `FvwmScript` nad elsewhere which aren't using `fxmalloc()` or `fxcalloc()`.  You could look at changing those, by writing a coccinelle script...

klebertarcisio:
Wow! It was too much of a coincidence. I looked at the last commit from the master and saw that the problem is solved. It's great!

ThomasAdam:
I usually spend Wednesday's on easy stuff -- this was just a coincidence.  :)

PR #156, opened 2020-07-02
Replace calloc by fxcalloc

klebertarcisio:
Replace calloc by fxcalloc

### DIFF
--- a/libs/FScreen.c
+++ b/libs/FScreen.c
@@ -83,7 +83,7 @@ monitor_new(void)
 {
 	struct monitor	*m;
 
-	m = calloc(1, sizeof *m);
+	m = fxcalloc(1, sizeof *m);
 	//m->si = calloc(1, sizeof(struct screen_info));
 
 	return (m);
--- a/libs/FScreen.c
+++ b/libs/FScreen.c
@@ -84,7 +84,6 @@ monitor_new(void)
 	struct monitor	*m;
 
 	m = fxcalloc(1, sizeof *m);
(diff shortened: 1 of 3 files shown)


libs/FScreen.c
```diff
@@ -83,7 +83,7 @@ monitor_new(void)
 {
 	struct monitor	*m;
 
-	m = calloc(1, sizeof *m);
+	m = fxcalloc(1, sizeof *m);
 	//m->si = calloc(1, sizeof(struct screen_info));
```
ThomasAdam, line 87: This change isn't correct, @klebertarcisio -- you need to update your copy of `master` in your repository first, then rebase this change on to it.

klebertarcisio: Hi @ThomasAdam, thanks for the review.
 
Why is this change incorrect? My master is up to date. In the last commit, you changed the function **screen_info_new**, but in this one I'm changing the function **monitor_new**. Do they both need changes?

ThomasAdam: Hi @klebertarcisio,

Ah, that's my mistake, sorry.  I had misread the change and was confused by the comment which doesn't need to be there, hence the following:

```suggestion
```

This change looks good.

ThomasAdam:
@klebertarcisio,

Have you been able to augment the coccinelle scripts for this detection of `calloc -> fxcalloc` and `malloc -> fxmalloc`?

klebertarcisio:
Hi @ThomasAdam. Sorry for the delay. No, I haven't used the script yet.

PR #157, opened 2020-07-03
Edit to Confirm Quit

lgsobalvarro:
Adds shutdown and reboot options


somiaj:
The issue with adding a shutdown/reboot to the FvwmScript-ConfirmQuit script is there is no universal way (that I know of) to ensure the user running the command has correct permissions to run shutdown or reboot the system. For instance this script won't work on my system, as I need to escalate privileges to shutdown or reboot the system.

Since it will be unknown if fvwm is running with sufficient permissions to be able to shutdown or restart the system (or what method such as sudo, systemctl, polkit, etc. is needed to escalate privileges) , I wouldn't add those options to the script (though this would make a good Wiki entry to those who want to expand the script to include such options).

lgsobalvarro:
I agree. I have being playing with it a bit more, and I don't see any simple way to ensure the command has correct permissions. Perhaps adding this to the Wiki as you suggest would be a better choice. However @thomas_adman suggested it could be possible to implement.

somiaj:
Keep in mind that fvwm is also not a DE, and as such it is okay that the default-config is missing features, and using the wiki to provide examples of common things is useful as one takes the default config and starts to tweak it to their liking. One of my visions of the wiki (tried this with the decor and panel examples) is to provide things that can be added onto the default-config (or anyone config) to allow users to pick and choose the features they want.

ThomasAdam:
OK.  Closing this for now.  Reopen this if there's a different solution you feel should be included.

PR #160, opened 2020-07-03
Avoid use uninitialized variables

klebertarcisio:
I think this patch avoids any possible use of an uninitialized variable.
The **get_icon_picture_geometry** and **get_icon_title_geometry** functions can return false. In this case, the variables **g.x** and **g.y** could be empty.

### DIFF
--- a/fvwm/events.c
+++ b/fvwm/events.c
@@ -417,23 +417,25 @@ static inline void _handle_cr_on_icon(XEvent *e, FvwmWindow *fw)
 	{
 		rectangle g;
 
-		get_icon_picture_geometry(fw, &g);
-		xwc.x = g.x;
-		xwc.y = g.y;
-		xwcm = cre->value_mask & (CWX | CWY);
-		XConfigureWindow(
-			dpy, FW_W_ICON_PIXMAP(fw), xwcm, &xwc);
+		if(get_icon_picture_geometry(fw, &g)){
+			xwc.x = g.x;
+			xwc.y = g.y;
+			xwcm = cre->value_mask & (CWX | CWY);
+			XConfigureWindow(
+				dpy, FW_W_ICON_PIXMAP(fw), xwcm, &xwc);
+		}
 	}
 	if (FW_W_ICON_TITLE(fw) != None)
 	{
 		rectangle g;
 
-		get_icon_title_geometry(fw, &g);
-		xwc.x = g.x;
-		xwc.y = g.y;
-		xwcm = cre->value_mask & (CWX | CWY);
-		XConfigureWindow(
-			dpy, FW_W_ICON_TITLE(fw), xwcm, &xwc);
+		if(get_icon_title_geometry(fw, &g)){
+			xwc.x = g.x;
+			xwc.y = g.y;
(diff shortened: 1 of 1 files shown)


fvwm/events.c
```diff
@@ -417,23 +417,25 @@ static inline void _handle_cr_on_icon(XEvent *e, FvwmWindow *fw)
...
-		xwcm = cre->value_mask & (CWX | CWY);
-		XConfigureWindow(
-			dpy, FW_W_ICON_PIXMAP(fw), xwcm, &xwc);
+		if(get_icon_picture_geometry(fw, &g)){
+			xwc.x = g.x;
+			xwc.y = g.y;
+			xwcm = cre->value_mask & (CWX | CWY);
+			XConfigureWindow(
+				dpy, FW_W_ICON_PIXMAP(fw), xwcm, &xwc);
+		}
```
ThomasAdam, line 426: So this change seems OK, however I think that:

* `get_icon_picture_geometry()` will call `memset()` on `g` in the case where there's nothing for that function to do (i.e., it returns `False`, so `xwc.?` will receive `0`.
* This change now *conditionally* only calls `XConfigureWindow()` if `get_icon_picture_geometry()` returns `True`.  This isn't correct, as many applications qill still want to receive this event, regardless of what `get_icon_picture_geometry()` returns.

Were you getting compiler warnings for this code?

fvwm/events.c
```diff
@@ -417,23 +417,25 @@ static inline void _handle_cr_on_icon(XEvent *e, FvwmWindow *fw)
...
-		xwcm = cre->value_mask & (CWX | CWY);
-		XConfigureWindow(
-			dpy, FW_W_ICON_TITLE(fw), xwcm, &xwc);
+		if(get_icon_title_geometry(fw, &g)){
+			xwc.x = g.x;
+			xwc.y = g.y;
+			xwcm = cre->value_mask & (CWX | CWY);
+			XConfigureWindow(
+				dpy, FW_W_ICON_TITLE(fw), xwcm, &xwc);
+		}
```
ThomasAdam, line 438: Same comments as above apply here.

fvwm/events.c
klebertarcisio: Hi @ThomasAdam, thanks for the review. No, I did not receive warnings from the compiler.
In this case, would the best change be just to replace the return of the **boolean** function to **void**?

ThomasAdam: Hi @klebertarcisio,

Heavens, no!  :)  I'm saying that there's no uninitialised variables -- `0` is perfectly legitimate here in the case when `False` is returned.  Hence, there's no modification required; I think this PR can be closed.

If you'd like me to suggest some things you could work on, do please let me know, I've got a list of things about a mile high...

klebertarcisio: If a function returns a value and that value is not received, I think this is an indication that the function should be void. However, you may be considering using this return in future cases. I got your point.

ThomasAdam:
Per review comments, I'm closing this PR.

Thanks, @klebertarcisio!

PR #161, opened 2020-07-05
update documentation for PrintInfo : it now writes to ~/fvwm3-output instead of stderr

d-e-e-p:
PrintInfo uses fvwm_debug instead of fprintf(stderr, ...

### DIFF
--- a/doc/commands/PrintInfo.xml
+++ b/doc/commands/PrintInfo.xml
@@ -23,7 +23,13 @@
 
 <para>Print information on
 <replaceable>subject</replaceable>
-on stderr.  An optional integer argument
+to debug log file, which defaults to <filename>$HOME/.fvwm/fvwm3-output.log</filename> . Environment variables <envar>$FVWM_USERDIR</envar> and 
+<envar>$FVWM3_LOGFILE</envar> can alter this default.  For this logfile to be 
+written, either fvwm3 has to be started with <option>-v</option> option or 
+<option>SIGUSR2</option> signal can be used to toggle opening/closing debug 
+log file.</para>
+
+<para> An optional integer argument
 <replaceable>verbose</replaceable>
 defines the level of information which is given.
 The current valid subjects are:</para>
--- a/doc/fvwm/description.xml
+++ b/doc/fvwm/description.xml
@@ -11,18 +11,14 @@ memory consumption, provide a 3D look to window frames, and a
 virtual desktop.</para>
 
 <para>Note that there are several window managers around that have
-"fvwm" in their name.  In the past, version 2.x of fvwm was
-commonly called fvwm2 to distinguish it from the former version
-1.x (fvwm or even fvwm1).  Since version 1.x has been replaced by
-version 2.x a long time ago we simply call version 2.x and all
-versions to come, fvwm, throughout this document, and the
-executable program is named fvwm.  There is an fvwm offspring
-called fvwm95, it is mostly a patched version of fvwm-2.0.43.  The
-main goal of fvwm95 was to supply a Windows 95 like look and
-feel. Since then, fvwm has been greatly enhanced and practically
-all fvwm95 features can be achieved by fvwm.</para>
+"fvwm" in their name. Fvwm3 is the successor to fvwm2, which 
+preceded the 1.x versions of fvwm. This version is simply called
+fvwm throughout this document, while the main executable is named 
+fvwm3.</para>
 
-<para>Fvwm provides both a large
+<para> Fvwm is intended to have a small memory footprint but a rich feature 
+set, be extremely customizable and extendible, and have a high degree of Motif 
+mwm compatibility.  Fvwm provides both a large
 <emphasis remap='I'>virtual desktop</emphasis>
 and
 <emphasis remap='I'>multiple disjoint desktops</emphasis>
--- a/doc/fvwm/environment.xml
+++ b/doc/fvwm/environment.xml
@@ -21,11 +21,31 @@ are the following:
     </listitem>
   </varlistentry>
   <varlistentry>
+    <term><envar>FVWM_USERDIR</envar></term>
+    <listitem>
+      <para>Used to determine the user's data directory for reading and
+	writing files. If this variable is not already
+	set, it is set by fvwm to <filename>$HOME/.fvwm</filename>, which
+	is the default user's data directory.</para>
+    </listitem>
+  </varlistentry>
+  <varlistentry>
     <term><envar>FVWM3_LOGFILE</envar></term>
     <listitem>
       <para>Used to determine the path and filename to log debug information
-	      from fvwm3.
+          from fvwm3. By default debug log is written to 
+          <envar>$FVWM_USERDIR</envar>/fvwm3-output.log . If an absolute path
+          is specified (starting with /) then <envar>$FVWM_USERDIR</envar> is 
+          ignored, otherwise the log is written to 
+          <envar>$FVWM_USERDIR</envar>/<envar>$FVWM3_LOGFILE</envar> .
       </para>
+    </listitem>
+  </varlistentry>
+  <varlistentry>
+    <term><envar>FVWM_DATADIR</envar></term>
+    <listitem>
+        <para>Set by fvwm to the directory containing fvwm config and module 
+            data.</para>
     </listitem>
   </varlistentry>
   <varlistentry>
@@ -33,15 +53,6 @@ are the following:
     <listitem>
       <para>Set by fvwm to the directory containing the standard fvwm
 	modules.</para>
-    </listitem>
-  </varlistentry>
-  <varlistentry>
-    <term><envar>FVWM_USERDIR</envar></term>
-    <listitem>
-      <para>Used to determine the user's data directory for reading and
-	sometimes writing personal files. If this variable is not already
-	set, it is set by fvwm to <filename>$HOME/.fvwm</filename>, which
-	is the default user's data directory.</para>
     </listitem>
   </varlistentry>
   <varlistentry>
@@ -70,4 +81,5 @@ and normally are removed automatically when not used anymore.</para>
     </listitem>
   </varlistentry>
 </variablelist>
+<para><command>fvwm-config --info</command> is useful to query fvwm3 compiled-in paths</para>
 </section>
--- a/doc/fvwm/options.xml
+++ b/doc/fvwm/options.xml
@@ -79,8 +79,7 @@ instead of the name obtained from the environment variable
   <listitem>
 <para>Puts X transactions in synchronous mode, which dramatically slows
 things down, but guarantees that fvwm's internal error messages
-are correct. Also causes fvwm to output debug messages while
-running.</para>
+are correct.</para>
   </listitem>
   </varlistentry>
   <varlistentry>
@@ -93,9 +92,8 @@ running.</para>
 <replaceable>config-file</replaceable>
 instead of
 <filename>~/.fvwm/config</filename>
-as its initialization file.  This is equivalent to
--c '<fvwmref cmd="Read"/>
-<replaceable>config-file</replaceable>'.</para>
+as its initialization file. <envar>$FVWM_USERDIR</envar> can also be used to 
+change location of default user directory <filename>~/.fvwm</filename>.</para> 
   </listitem>
   </varlistentry>
   <varlistentry>
@@ -234,7 +232,7 @@ follow:</para>
   </varlistentry>
 </variablelist>
 
-<para>These defaults may change before version 2.6. Note that if
+<para>Note that if
 you use a private color map (i.e., fvwm is started with the
 <option>-C</option> or the <option>-I</option> options), then
 other defaults are used.</para>
@@ -370,7 +368,22 @@ internal debugging and should only be used by developers.</para>
   <varlistentry>
   <term><option>-v</option></term>
   <listitem>
-<para>Enables debug logging.  Write to ~/fvwm3-output.log in append mode.</para>
+<para>Enables debug logging.  Writes in append mode to fvwm log file, 
+which is ~/.fvwm/fvwm3-output.log by default. See ENVIRONMENT section on 
+how to override this location on fvwm3 startup using 
+<envar>$FVWM_USERDIR</envar> or <envar>$FVWM3_LOGFILE</envar> .</para>
+  <variablelist>
+    <varlistentry>
+      <term>Logging can also be dynamically toggled on and off using signals:</term>
+      <listitem>
+        <simplelist>
+      <member><option>SIGUSR1</option> : used as a signal to restart Fvwm</member>
+      <member><option>SIGUSR2</option> : used as a signal to toggle opening/closing debug log file</member>
+        </simplelist>
+      </listitem>
+    </varlistentry>
+  </variablelist>
+
   </listitem>
   </varlistentry>
 </variablelist>
--- a/libs/log.c
+++ b/libs/log.c
@@ -58,7 +58,7 @@ log_open(const char *fvwm_userdir)
 		expanded_path = expand_path(path);
 
 		path = fxstrdup(expanded_path);
-		if (strchr(expanded_path, '/') != NULL)
+		if (expanded_path[0] == '/')
 		{
 				path = fxstrdup(expanded_path);
 		} else {
--- a/libs/log.c
+++ b/libs/log.c
@@ -45,9 +45,10 @@ log_set_level(int ll)
 
 /* Open logging to file. */
 void
-log_open(void)
+log_open(const char *fvwm_userdir)
 {
-	char		*path = fxstrdup(FVWM3_LOGFILE_DEFAULT);
+	char *path;
+	xasprintf(&path, "%s/%s", fvwm_userdir, FVWM3_LOGFILE_DEFAULT);
 
 	if (getenv("FVWM3_LOGFILE")) {
 		const char	*expanded_path;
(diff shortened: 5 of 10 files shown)


d-e-e-p:
I think for most mortals the SIGUSR2 way is going to be confusing--especially for PrintInfo.                                      Ideally PrintInfo internally turns on logging if it wasn't on before, and then restores state after running. eg:    
                                                                                                                                  
  ```
                                                                                                                                
[-l log-file] changes the location of fvwm default log file from ~/fvwm3-output.log to log-file                                   
[-v ]         Enables debug logging.  Writes in append mode to fvwm log file, which is ~/fvwm3-output.log by default              
              Logging can also be dynamically turned on and off by sending SIGUSR2 signal to fvwm with "kill -USR2 pid"           
                                                                                                                                  
PrintInfo subject [verbose]                                                                                                       
                                                                                                                                  
    Print information on subject to fvwm log file, which is ~/fvwm3-output.log by default.                                        
    ...                                                                                                                           
                 
```        

I have "PrintInfo ImageCache 1" in my startup config file somewhere...

**Comment by ThomasAdam** [comment]
Hi @d-e-e-p,

Well, we're not talking about most mere mortals, this is a means of independently capturing useful information.

SIGUSR2 is tremendously useful for diagnostics.

I'm not going to make PrintInfo a special case for weird toggling either.

Let me know if you plan on working on this further. If not, I'll add a note to do it.

Cheers,
Thomas

**Comment by d-e-e-p** [comment]
Is there an updated doc/fvwm/options.xml somewhere? in the current version it seems -l is aliased to --color-limit .
it seems your plan is:

```
                                                                                                                                
[-l log-file] changes the location of fvwm default log file from ~/fvwm3-output.log to log-file                                   
[-v ]         Enables debug logging.  Writes in append mode to fvwm log file, which is ~/fvwm3-output.log by default              
              Logging can also be dynamically toggled on and off by sending SIGUSR2 signal to fvwm with "kill -USR2 pid"          
                                                                                                                                  
PrintInfo subject [verbose]                                                                                                       
                                                                                                                                  
  Print information on subject to fvwm log file, which is ~/fvwm3-output.log by default.                                          
  PrintInfo writes this only when debugging is enabled, either by invoking fvwm3 with -v option or sending fvwm3 SIGUSR2 signal.                              
  ...   
```                                                                                                 

I could help redo the doc if you need the help.

**Comment by ThomasAdam** [comment]
Hi @d-e-e-p,

Ah -- thanks.  You've spotted a bug there, since `-l` only works due to the order the options are parsed in, so we should remove `-l` as an alias to `--color-limit` since it's not even used.  Care to do that, too?

**Comment by d-e-e-p** [comment]
i'll take a stab. things to do in options.xml :

- [ ] remove -l and -L options, but leave intact all that dated discussion of XFree-4.2 color allocation and netscape. ..
- [ ] add -l option that changes default debug logfile
- [ ] add info about SIGUSR1/SIGUSR2

**Comment by d-e-e-p** [comment]
if both -l option and FVWM3_LOGFILE env var is specified, I presume -l wins?  or should we remove one of these...
other random ttd:

- [ ] link rot eg http://www.freedesktop.org/wiki/Standards_2fwm_2dspec to https://specifications.freedesktop.org/wm-spec/wm-spec-1.3.html                            
- [ ] remove references to missing modules, eg  FvwmM4/FvwmCpp
- [ ] remove obsolete commands (  ColorLimit,  IconPath ... )
- [ ] add DesktopConfiguration

**Comment by ThomasAdam** [comment]
Hi @d-e-e-p,

Don't worry about `ColorLimit` and `IconPath` for now, as the code paths still shim to the newer implementations.

As for `DesktopConfiguration`, there's already a section in `man fvwm3` about that.

**Comment by ThomasAdam** [comment]
Hi @d-e-e-p,

I'd misread something you mentioned earlier.  I thought you were saying that I had previously added `-l` to change the log path/filename, and that this was in conflict with `--color-limit`.  But that's not what you meant.  Apologies for the confusion.

I don't necessarily think we need `-l`, when we already have `FVWM3_LOGFILE` as an env var check...  At this point, it's all a variation on the same theme.

**Comment by somiaj** [comment]
Hi @d-e-e-p 

Seeing this pull request caused me to notice that the log file was put in $HOME instead of $FVWM_USERDIR by default. I think it would probably be best to use $FVWM_USERDIR to keep all fvwm files in a central location. Here is a patch that prepends $FVWM_USERDIR to the default log file location and to $FVWM3_LOGFILE if it doesn't contain a '/'. Since you are working on this already, here is the patch that you can choose to include with your changes or not.

```diff
diff --git a/fvwm/fvwm3.c b/fvwm/fvwm3.c
index ff4f9a7a..a03b55ae 100644
--- a/fvwm/fvwm3.c
+++ b/fvwm/fvwm3.c
@@ -295,7 +295,7 @@ Restart(int sig)
 static RETSIGTYPE
 ToggleLogging(int sig)
 {
-       log_toggle();
+       log_toggle(fvwm_userdir);

        SIGNAL_RETURN;
 }
@@ -2069,7 +2069,7 @@ int main(int argc, char **argv)
                else if (strcmp(argv[i], "-v") == 0)
                {
                        log_set_level(1);
-                       log_open();
+                       log_open(fvwm_userdir);
                }
                else
                {
diff --git a/libs/log.c b/libs/log.c
index 58fcdccd..81da555e 100644
--- a/libs/log.c
+++ b/libs/log.c
@@ -45,9 +45,10 @@ log_set_level(int ll)

 /* Open logging to file. */
 void
-log_open(void)
+log_open(const char *fvwm_userdir)
 {
-       char            *path = fxstrdup(FVWM3_LOGFILE_DEFAULT);
+       char *path;
+       asprintf(&path, "%s/%s", fvwm_userdir, FVWM3_LOGFILE_DEFAULT);

        if (getenv("FVWM3_LOGFILE")) {
                const char      *expanded_path;
@@ -56,7 +57,12 @@ log_open(void)
                path = getenv("FVWM3_LOGFILE");
                expanded_path = expand_path(path);

-               path = fxstrdup(expanded_path);
+               if (strchr(expanded_path, "/") != NULL)
+               {
+                       path = fxstrdup(expanded_path);
+               } else {
+                       asprintf(&path, "%s/%s", fvwm_userdir, expanded_path);
+               }
                free((char *)expanded_path);
        }

@@ -77,11 +83,11 @@ log_open(void)

 /* Toggle logging. */
 void
-log_toggle(void)
+log_toggle(const char *fvwm_userdir)
 {
        if (log_level == 0) {
                log_level = 1;
-               log_open();
+               log_open(fvwm_userdir);
                fvwm_debug(NULL, "log opened (because of SIGUSR2)");
        } else {
                fvwm_debug(NULL, "log closed (because of SIGUSR2)");
diff --git a/libs/log.h b/libs/log.h
index c43347c8..7933e9e5 100644
--- a/libs/log.h
+++ b/libs/log.h
@@ -6,8 +6,8 @@

 void   log_set_level(int);
 int    log_get_level(void);
-void   log_open(void);
-void   log_toggle(void);
+void   log_open(const char *);
+void   log_toggle(const char *);
 void   log_close(void);
 void printflike(2, 3) fvwm_debug(const char *, const char *, ...);
```

doc/commands/PrintInfo.xml
```diff
@@ -23,7 +23,7 @@
 
 <para>Print information on
 <replaceable>subject</replaceable>
-on stderr.  An optional integer argument
+to ~/fvwm3-output.log. fvwm3 has to be run with -v option for this log file to be written.  An optional integer argument
```
ThomasAdam:
```suggestion
to ~/fvwm3-output.log. fvwm3 has to be run with -v option for this log file to be written, or sent SIGUSR2.  An optional integer argument
```

doc/fvwm/description.xml
```diff
@@ -11,18 +11,14 @@ memory consumption, provide a 3D look to window frames, and a
...
-versions to come, fvwm, throughout this document, and the
-executable program is named fvwm.  There is an fvwm offspring
-called fvwm95, it is mostly a patched version of fvwm-2.0.43.  The
-main goal of fvwm95 was to supply a Windows 95 like look and
-feel. Since then, fvwm has been greatly enhanced and practically
-all fvwm95 features can be achieved by fvwm.</para>
+"fvwm" in their name. Fvwm3 is the successor to fvwm2, which 
+preceded the 1.x versions of fvwm. This version is simply called
+fvwm throughout this document, while the main executable is named 
+fvwm3.</para>
```
ThomasAdam, line 17: I'm OK with this, but we should standardise on `Fvwm3` in the future.

doc/fvwm/options.xml
```diff
@@ -93,8 +92,9 @@ running.</para>
 <replaceable>config-file</replaceable>
 instead of
 <filename>~/.fvwm/config</filename>
-as its initialization file.  This is equivalent to
--c '<fvwmref cmd="Read"/>
+as its initialization file. <envar>$FVWM_USERDIR</envar> can also be used to 
+change location of default user directory <filename>~/.fvwm</filename>. Another
+option is to use -c '<fvwmref cmd="Read"/>
```
ThomasAdam, line 97: I'm not in favour of advocating `-c` -- it's one of those options which will go...

doc/fvwm/options.xml
```diff
@@ -370,7 +370,23 @@ internal debugging and should only be used by developers.</para>
   <varlistentry>
   <term><option>-v</option></term>
   <listitem>
-<para>Enables debug logging.  Write to ~/fvwm3-output.log in append mode.</para>
+<para>Enables debug logging.  Writes in append mode to fvwm log file, 
+which is ~/.fvwm/fvwm3-output.log by default. See ENVIRONMENT section on 
+how to override this location using <envar>$FVWM_USERDIR</envar> or 
+<envar>$FVWM3_LOGFILE</envar>.</para>
+
```
ThomasAdam, line 377: Again, `FVWM_USERDIR` is set once, and really shouldn't be updated once `Fvwm` is running.

d-e-e-p:
Tested absolute and relative path settings of FVWM_LOGFILE, and also using the FVWM_USERDIR to override logfile location.
These three options seem to work....

doc/fvwm/environment.xml
```diff
@@ -20,28 +20,39 @@ are the following:
...
+    </listitem>
+  </varlistentry>
   <varlistentry>
     <term><envar>FVWM3_LOGFILE</envar></term>
     <listitem>
       <para>Used to determine the path and filename to log debug information
-	      from fvwm3.
+          from fvwm3. By default debug log is written to 
+          <envar>$FVWM_USERDIR</envar>/fvwm3-output.log . If an absolute path
+          is specified (starting with /) then <envar>$FVWM_USERDIR</envar> is
```
somiaj, line 38: Currently the code is 'contains /' not 'starting with /'

d-e-e-p:
@somiaj I would vote for path names that contain but not start with slash to be handled as relative paths. So if `FVWM3_LOGFILE` is set to `logs/fvwm.log` and `FVWM_USERDIR` is not set, log is written to `~/.fvwm/logs/fvwm.log` :

Perhaps long term make userdir act the same way?

```
userdir starts_with_slash  $FVWM_USERDIR
userdir contains_slash     $HOME/$FVWM_USERDIR
userdir no_slash           $HOME/$FVWM_USERDIR

logfile starts_with_slash  $FVWM3_LOGFILE
logfile contains_slash     $FVWM_USERDIR/$FVWM3_LOGFILE
logfile no_slash           $FVWM_USERDIR/$FVWM3_LOGFILE

```

ThomasAdam:
Hi @d-e-e-p,

Interesting thought, but I think this is starting to become over engineered.

The whole point of this is for diagnostics. I really don't want anything more complicated or involved than the patch included in this PR provides. It's such a tiny thing that is rarely going to be used.

We've plenty of other items to turn our attention to yet.

libs/log.c
```diff
@@ -57,6 +58,12 @@ log_open(void)
 		expanded_path = expand_path(path);
 
 		path = fxstrdup(expanded_path);
+		if (strchr(expanded_path, '/') != NULL)
```
somiaj, line 61: To make the patch for $FVWM3_LOGFILE use 'starts with /' instead of 'contains /', change this line to (I haven't tested this, so hopefully my C syntax is correct):

```suggestion
		if (expanded_path[0] == '/')
```

I think it is safe to assume expanded_path is not NULL so just checking if the first character is '/' should be sufficient. This small change should make the logic starts with '/' vs contains to prepend $FVWM_USERDIR or not. So update this if you want that change.

If $FVWM_USERDIR prepends $HOME or not is a separate issue (if it even is one -- I think the logic there is currently fine) and shouldn't be part of this patch IMO.

libs/log.c
```diff
@@ -57,6 +58,12 @@ log_open(void)
 		expanded_path = expand_path(path);
 
 		path = fxstrdup(expanded_path);
+		if (expanded_path[0] == '/')
+		{
+				path = fxstrdup(expanded_path);
```
ThomasAdam, line 63: The logic here is incorrect (not to mention this is now causing a memory leak).

We set `path` to `expanded` path, and then set `path` to `expanded_path` again if `expanded_path` begins with a `/`.

This isn't right.  Can someone please fix this logic?

libs/log.c
```diff
@@ -57,6 +58,12 @@ log_open(void)
 		expanded_path = expand_path(path);
 
 		path = fxstrdup(expanded_path);
```
somiaj, line 60: Not sure why this line is here, it should have been deleted (and only used if starts with /).

```suggestion
```

ThomasAdam: Better, but we need to do a bit more, as in:

```
diff --git a/libs/log.c b/libs/log.c
index 23a65a70c..977c13c51 100644
--- a/libs/log.c
+++ b/libs/log.c
@@ -47,23 +47,21 @@ log_set_level(int ll)
 void
 log_open(const char *fvwm_userdir)
 {
-       char *path;
+       char *path, *logfile_env;
+
        xasprintf(&path, "%s/%s", fvwm_userdir, FVWM3_LOGFILE_DEFAULT);
 
-       if (getenv("FVWM3_LOGFILE")) {
+       logfile_env = getenv("FVWM3_LOGFILE");
+       if (logfile_env != NULL) {
                const char      *expanded_path;
 
                free(path);
-               path = getenv("FVWM3_LOGFILE");
-               expanded_path = expand_path(path);
+               expanded_path = expand_path(logfile_env);
 
-               path = fxstrdup(expanded_path);
                if (expanded_path[0] == '/')
-               {
-                               path = fxstrdup(expanded_path);
-               } else {
-                               xasprintf(&path, "%s/%s", fvwm_userdir, expanded_path);
-               }
+                       path = fxstrdup(expanded_path);
+               else
+                       xasprintf(&path, "%s/%s", fvwm_userdir, expanded_path);
                free((char *)expanded_path);
        }
```

ThomasAdam: Additionally, this code change in its entirety needs to be split off as a separate commit, and not just shoe-horned in to the other commits as it's nothing to do with these on-going documentation changes.

@d-e-e-p -- are you OK doing that?

d-e-e-p:
Ok, so plan is to do a separate pull request handling FVWM3_LOGFILE .. perhaps starting with version where document changes have been merged in?

ThomasAdam:
Hi @d-e-e-p,

If you separate them out, I'll sort out the order I want to merge them in.

d-e-e-p:
I'm not sure how to do multiple fork and pull requests--the 2 repos are at:

```
git request-pull origin/master https://github.com/d-e-e-p/fvwm3_code
git request-pull origin/master https://github.com/d-e-e-p/fvwm3_doc
```

ThomasAdam:
Hi @d-e-e-p,

Thanks for separating those out, but that has a certain overhead which I didn't have in mind....

What I've done instead is taken your commits from this PR, and for the original commit which had both documentation and code changes, I ran:

```
git rebase -i master
```

Chose the commit which had the code changes with documentation, selected to 'edit' the commit, and then ran:

```
git checkout HEAD~1
```

To ammend the commit in question, split out the files to separate commits, and continued the rebase.

I then squashed your documentation commits together, and updated the code changes with my additional patch.

The result of all of this is here:  `ta/deep-docs`

@d-e-e-p -- you should be able to checkout that branch and change as necessary, and update your PR here with what's there.  But I think what's there looks OK.  But do please check it over.

d-e-e-p:
ta/deep-docs looks good. tested the env var settings with following script.
(completely unrelated, lib**n**son in configure.ac should be lib**b**son)

```
#!/bin/sh -x

unset FVWM_USERDIR FVWM3_LOGFILE


rm -rf /tmp/flog
mkdir  /tmp/flog

rm -f $HOME/.fvwm/fvwm3-output.log                                              
timeout 10s fvwm3 -r -D -v  >> debug.log  2>&1
ls -l $HOME/.fvwm/fvwm3-output.log

export FVWM_USERDIR=/tmp/flog
timeout 10s fvwm3 -r -D -v  >> debug.log  2>&1
ls -l $FVWM_USERDIR/fvwm3-output.log

export FVWM3_LOGFILE=case1.log
timeout 10s fvwm3 -r -D -v  >> debug.log  2>&1
ls -l $FVWM_USERDIR/$FVWM3_LOGFILE

mkdir -p  /tmp/flog/log/log
export FVWM3_LOGFILE=log/log/case2.log
timeout 10s fvwm3 -r -D -v  >> debug.log  2>&1
ls -l $FVWM_USERDIR/$FVWM3_LOGFILE

export FVWM3_LOGFILE=/tmp/flog/case3.log
timeout 10s fvwm3 -r -D -v  >> debug.log  2>&1
ls -l $FVWM3_LOGFILE 

echo "done"

```

ThomasAdam:
Hi @d-e-e-p,

Thanks -- will look at merging this soon, unless you have anything else you want to change?

As for libbson spelling, see:  https://github.com/fvwmorg/fvwm3/issues/162

Do you think you

ThomasAdam:
Thanks all.  I've merged `ta/deep-docs` to master now, so I'm closing this.

Next time, let's try and keep these commits small and relevant to the specific changes being made.

PR #176, opened 2020-07-28
Make MFL optional

tuncer:
MFL pulls in libevent and libbson while also, arguably, not being a core
module.  Therefore, we build a stub implementation if dependencies are
unavailable. It could make sense to also expose an autoconf flag.


ThomasAdam:
Hi @tuncer,

Thank you for your patch. However, it's a module only for how it's written. It's by no means optional.

Kindly,
Thomas

PR #181, opened 2020-08-12
making menuitem text always vertically centered.

mikeandmore:
Currently, if a menuitem contains a miniicon, then the text isn't vertically centered. This patch fixes this strange behaviour.

### DIFF
--- a/fvwm/menuitem.c
+++ b/fvwm/menuitem.c
@@ -429,7 +429,7 @@ void menuitem_paint(
 				y = MI_MINI_ICON(mi)[i]->height;
 			}
 		}
-		y -= font->height;
+		y -= font->height + ST_ITEM_GAP_ABOVE(ms) + ST_ITEM_GAP_BELOW(ms);
 		if (y > 1)
 		{
 			text_y += y / 2;


ThomasAdam:
This should take into account the size of the miniicon.

mikeandmore:
@ThomasAdam `calc_normal_item_height()` seems to suggest the menu item looks like the following:

```
------------------------------------------
|         gap(VerticalItemSpacing)
|-----------------------------------------
|         text
|-----------------------------------------
|         gap(VerticalItemSpacing)
------------------------------------------
```
No matter where you put the miniicon, the text position is always right under the gap.

Is that right?

mikeandmore:
I think I was wrong before. Yes, the height is chosen using the max of miniicon and the text + gap size.

The latest commit would fix the issue I'm encountering. The real issue is: if ItemSpacing is too large, then the text won't appear at the center.

I also tried ItemSpacing with 0 using this patch, it works well too.

ThomasAdam:
Hi @mikeandmore,

This looks much better, thanks.  Last thing left to do is to expand on your commit message, please, to explain *why* we're making this change, and what this is fixing as a result, then I'll merge this.

mikeandmore:
How about now?

ThomasAdam (approved):

PR #183, opened 2020-08-16
FvwmScript Comments Translation

phileimer:
Comments translated from French to English.


ThomasAdam:
Merci beaucoup, @phileimer!

PR #184, opened 2020-08-17
FvwmForm: change Timeout resolution to 100ms, instead of 1s

phileimer:
Old patch for version 2.6, merged with no modification in 3.0.
Needs testing.


ThomasAdam:
Thanks, @phileimer -- let's see if anyone does test this, and we can fix any mistakes as they crop up.

PR #214, opened 2020-09-08
Respect environment variable for MFL socket path

NsCDE:
Hi Thomas,

While reading man page for FvwmMFL, I saw reference to FVWMMFL_SOCKET, but this doesn't work because it was still in plan in modules/FvwmMFL/FvwmMFL.c.

I played a bit this evening and got this to work for me. Please see if this is ok, and if it can be merged. Thanks. :-)

### DIFF
--- a/modules/FvwmMFL/FvwmMFL.c
+++ b/modules/FvwmMFL/FvwmMFL.c
@@ -15,6 +15,7 @@
 #include "libs/queue.h"
 #include "libs/fvwmsignal.h"
 #include "libs/vpacket.h"
+#include "libs/getpwuid.h"
 
 #include <sys/socket.h>
 #include <sys/un.h>
@@ -40,8 +41,8 @@
 #include <event2/listener.h>
 #include <event2/util.h>
 
-/* XXX - should also be configurable via getenv() */
-#define SOCK "/tmp/fvwm_mfl.sock"
+/* This is also configurable via getenv() - see FvwmMFL(1) */
+#define MFL_SOCKET_DEFAULT "/tmp/fvwm_mfl.sock"
 #define MYNAME "FvwmMFL"
 
 static int debug;
@@ -113,6 +114,7 @@ static struct fvwm_msg *fvwm_msg_new(void);
 static void fvwm_msg_free(struct fvwm_msg *);
 static void register_interest(void);
 static void send_version_info(struct client *);
+static char *set_socket_path();
 
 static struct fvwm_msg *
 fvwm_msg_new(void)
@@ -134,8 +136,11 @@ fvwm_msg_free(struct fvwm_msg *fm)
 static void
 HandleTerminate(int fd, short what, void *arg)
 {
+	char *sock_path;
+
+	sock_path = set_socket_path();
 	fprintf(stderr, "%s: dying...\n", __func__);
-	unlink(SOCK);
+	unlink(sock_path);
 	fvwmSetTerminate(fd);
 }
 
@@ -623,15 +628,42 @@ static void
 fvwm_read(int efd, short ev, void *data)
 {
 	FvwmPacket	*packet;
+	char *sock_path;
+
+	sock_path = set_socket_path();
 
 	if ((packet = ReadFvwmPacket(efd)) == NULL) {
 		if (debug)
 			fprintf(stderr, "Couldn't read from FVWM - exiting.\n");
-		unlink(SOCK);
+		unlink(sock_path);
 		exit(0);
 	}
 
 	broadcast_to_client(packet);
+}
+
+static char *set_socket_path()
+{
+	char *mflsock_env, *sock_path;
+
+	/* Figure out if we are using default MFL socket path or we should respect */
+	/* environment variable FVWMMFL_SOCKET for FvwmMFL socket path */
+	xasprintf(&sock_path, "%s", MFL_SOCKET_DEFAULT);
+
+	mflsock_env = getenv("FVWMMFL_SOCKET");
+	if (mflsock_env != NULL) {
+		const char      *unrolled_path;
+
+		unrolled_path = expand_path(mflsock_env);
+
+		if (unrolled_path[0] == '/')
+			sock_path = fxstrdup(unrolled_path);
+		else
+			xasprintf(&sock_path, "%s/%s", getenv("FVWM_USERDIR"), unrolled_path);
+
+		free((char *)unrolled_path);
+	}
+	return (sock_path);
 }
 
 int main(int argc, char **argv)
@@ -639,6 +671,8 @@ int main(int argc, char **argv)
 	struct event_base     *base;
 	struct evconnlistener *fmd_cfd;
 	struct sockaddr_un    sin;
+
+	char *sock_path;
 
 	TAILQ_INIT(&clientq);
 
@@ -654,9 +688,11 @@ int main(int argc, char **argv)
 	}
 	setup_signal_handlers(base);
 
+	sock_path = set_socket_path();
+
 	memset(&sin, 0, sizeof(sin));
 	sin.sun_family = AF_LOCAL;
-	strcpy(sin.sun_path, SOCK);
+	strcpy(sin.sun_path, sock_path);
 
 	/* Create a new listener */
 	fmd_cfd = evconnlistener_new_bind(base, accept_conn_cb, NULL,
@@ -685,7 +721,7 @@ int main(int argc, char **argv)
 	SendFinishedStartupNotification(fc.fd);
 
 	event_base_dispatch(base);
-	unlink(SOCK);
+	unlink(sock_path);
 
 	return (0);
 }
--- a/modules/FvwmMFL/FvwmMFL.c
+++ b/modules/FvwmMFL/FvwmMFL.c
@@ -690,7 +690,6 @@ int main(int argc, char **argv)
 	memset(&sin, 0, sizeof(sin));
 	sin.sun_family = AF_LOCAL;
 	strcpy(sin.sun_path, sock_pathname);
-	free(sock_pathname);
 
 	/* Create a new listener */
 	fmd_cfd = evconnlistener_new_bind(base, accept_conn_cb, NULL,
--- a/modules/FvwmMFL/FvwmMFL.c
+++ b/modules/FvwmMFL/FvwmMFL.c
@@ -137,9 +137,6 @@ fvwm_msg_free(struct fvwm_msg *fm)
 static void
 HandleTerminate(int fd, short what, void *arg)
 {
-
-	sock_pathname = set_socket_pathname();
-
 	fprintf(stderr, "%s: dying...\n", __func__);
 	unlink(sock_pathname);
 	fvwmSetTerminate(fd);
@@ -629,7 +626,6 @@ static void
 fvwm_read(int efd, short ev, void *data)
 {
 	FvwmPacket	*packet;
-	sock_pathname = set_socket_pathname();
 
 	if ((packet = ReadFvwmPacket(efd)) == NULL) {
 		if (debug)
@@ -682,14 +678,14 @@ int main(int argc, char **argv)
 		return (1);
 	}
 
+	sock_pathname = set_socket_pathname();
+
 	/* Create new event base */
 	if ((base = event_base_new()) == NULL) {
 		fprintf(stderr, "Couldn't start libevent\n");
 		return (1);
 	}
 	setup_signal_handlers(base);
-
-	sock_pathname = set_socket_pathname();
 
 	memset(&sin, 0, sizeof(sin));
 	sin.sun_family = AF_LOCAL;
--- a/modules/FvwmMFL/FvwmMFL.c
+++ b/modules/FvwmMFL/FvwmMFL.c
@@ -47,7 +47,7 @@
 
 static int debug;
 struct fvwm_msg;
-static const char *sock_pathname;
+static char *sock_pathname;
 
 struct client {
 	struct bufferevent	*comms;
@@ -115,7 +115,7 @@ static struct fvwm_msg *fvwm_msg_new(void);
 static void fvwm_msg_free(struct fvwm_msg *);
 static void register_interest(void);
 static void send_version_info(struct client *);
-static const char *set_socket_pathname(void);
+static char *set_socket_pathname(void);
 
 static struct fvwm_msg *
 fvwm_msg_new(void)
@@ -139,6 +139,7 @@ HandleTerminate(int fd, short what, void *arg)
 {
 
 	sock_pathname = set_socket_pathname();
+
 	fprintf(stderr, "%s: dying...\n", __func__);
 	unlink(sock_pathname);
 	fvwmSetTerminate(fd);
@@ -628,7 +629,6 @@ static void
 fvwm_read(int efd, short ev, void *data)
 {
 	FvwmPacket	*packet;
-
 	sock_pathname = set_socket_pathname();
 
 	if ((packet = ReadFvwmPacket(efd)) == NULL) {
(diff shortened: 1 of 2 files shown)


modules/FvwmMFL/FvwmMFL.c
```diff
@@ -113,6 +114,7 @@ static struct fvwm_msg *fvwm_msg_new(void);
 static void fvwm_msg_free(struct fvwm_msg *);
 static void register_interest(void);
 static void send_version_info(struct client *);
+static char *set_socket_path();
```
ThomasAdam, line 117:
```suggestion
static char *set_socket_path(void);
```

This at least makes it >C89 compliant.  ;)

modules/FvwmMFL/FvwmMFL.c
```diff
@@ -134,8 +136,11 @@ fvwm_msg_free(struct fvwm_msg *fm)
 static void
 HandleTerminate(int fd, short what, void *arg)
 {
+	char *sock_path;
+
+	sock_path = set_socket_path();
 	fprintf(stderr, "%s: dying...\n", __func__);
-	unlink(SOCK);
+	unlink(sock_path);
```
ThomasAdam, line 143: We should only be setting `sock_path` once.  More on this later...

modules/FvwmMFL/FvwmMFL.c
```diff
@@ -623,23 +628,52 @@ static void
...
 			fprintf(stderr, "Couldn't read from FVWM - exiting.\n");
-		unlink(SOCK);
+		unlink(sock_path);
 		exit(0);
 	}
 
 	broadcast_to_client(packet);
 }
 
+static char *set_socket_path()
```
ThomasAdam, line 645:
```suggestion
static const char *
set_socket_path(void)
```

modules/FvwmMFL/FvwmMFL.c
```diff
@@ -623,23 +628,52 @@ static void
...
+{
+	char *mflsock_env, *sock_path;
+
+	/* Figure out if we are using default MFL socket path or we should respect */
+	/* environment variable FVWMMFL_SOCKET for FvwmMFL socket path */
+	xasprintf(&sock_path, "%s", MFL_SOCKET_DEFAULT);
+
+	mflsock_env = getenv("FVWMMFL_SOCKET");
+	if (mflsock_env != NULL) {
```
ThomasAdam, line 654: I think it would be easier to check for NULL first of all, as in:

```
mflsock_env = getenv("FVWMMFL_SOCKET");
if (mflsock_env == NULL) {
        return (xstrdup(MFL_SOCKET_DEFAULT));
}

unrolled_path = expand_path(mflsock_env);
if (unrolled_path[0] == '/')
    sock_path = fxstrdup(unrolled_path);
else
    xasprintf(&sock_path, "%s/%s", getenv("FVWM_USERDIR"), unrolled_path);
```

modules/FvwmMFL/FvwmMFL.c
```diff
@@ -623,23 +628,52 @@ static void
...
+		const char      *unrolled_path;
+
+		unrolled_path = expand_path(mflsock_env);
+
+		if (unrolled_path[0] == '/')
+			sock_path = fxstrdup(unrolled_path);
+		else
+			xasprintf(&sock_path, "%s/%s", getenv("FVWM_USERDIR"), unrolled_path);
+
+		free((char *)unrolled_path);
```
ThomasAdam, line 664:
```suggestion
		free((void *)unrolled_path);
```

modules/FvwmMFL/FvwmMFL.c
```diff
@@ -623,23 +628,52 @@ static void
...
+	return (sock_path);
+}
+
 int main(int argc, char **argv)
 {
 	struct event_base     *base;
 	struct evconnlistener *fmd_cfd;
 	struct sockaddr_un    sin;
 
+	char *sock_path;
```
ThomasAdam, line 675: Suggest making this a static variable at the top of this file, outside of `main()`

modules/FvwmMFL/FvwmMFL.c
```diff
@@ -654,9 +688,11 @@ int main(int argc, char **argv)
 	}
 	setup_signal_handlers(base);
 
+	sock_path = set_socket_path();
+
 	memset(&sin, 0, sizeof(sin));
 	sin.sun_family = AF_LOCAL;
-	strcpy(sin.sun_path, SOCK);
+	strcpy(sin.sun_path, sock_path);
```
ThomasAdam, line 695: Memory leak...

```
free(sock_path)
```

ThomasAdam (commented):
Hi @NsCDE,

Thanks for this.  I clearly got ahead of myself in documenting this feature, despite never having written the code.  Ooops!

I've left some comments in-line for you to ponder.

I think the biggest suggestion I've got is to make `const char *sock_path` a global by declaring it as: `static const char *sock_path`, putting it someone near the top of the file.  This isn't going to change/be read again, once `FvwmMFL` has been started.  So we only need to work it out once.

I've also suggested a minor change in logic for `set_sock_path()` -- perhaps this now needs to be renamed since it's not just setting the path, it's also the path to the socket as well?  I'll leave that up to you to decide if it's worth while.

Thanks for this though!

modules/FvwmMFL/FvwmMFL.c
NsCDE: Hi Thomas,

As you pointed, it is a full filesystem address, so I have renamed sock_path into sock_pathname, made freeing of variable where you pointed it out, *sock_pathname is now global and at the top with other similar variables, but I have one problem:

I have changed everything you pointed out, but if I make "const" socket_pathname and put const function set_socket_pathname prototype and body, code complains but xasprintf complains (see below). I don't know how to resolve this except of not making this "const".

  CC       FvwmMFL.o
FvwmMFL.c: In function ‘set_socket_pathname’:
FvwmMFL.c:662:16: warning: passing argument 1 of ‘xasprintf’ from incompatible pointer type [-Wincompatible-pointer-types]
  662 |      xasprintf(&sock_pathname, "%s/%s", getenv("FVWM_USERDIR"), unrolled_path);
      |                ^~~~~~~~~~~~~~
      |                |
      |                const char **
In file included from ../../libs/fvwmlib.h:13,
                 from ../../libs/Flocale.h:14,
                 from ../../fvwm/fvwm.h:43,
                 from FvwmMFL.c:11:
../../libs/safemalloc.h:7:16: note: expected ‘char **’ but argument is of type ‘const char **’
    7 | int  xasprintf(char **, const char *, ...);
      |                ^~~~~~~
FvwmMFL.c: In function ‘main’:
FvwmMFL.c:694:7: warning: passing argument 1 of ‘free’ discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
  694 |  free(sock_pathname);
      |       ^~~~~~~~~~~~~
In file included from ../../config.h:474,
                 from FvwmMFL.c:9:
/usr/include/stdlib.h:565:25: note: expected ‘void *’ but argument is of type ‘const char *’
  565 | extern void free (void *__ptr) __THROW;
      |                   ~~~~~~^~~~~
  CCLD     FvwmMFL

NsCDE:
Hi. I just made proposed corrections, and I have problems with const + xasprintf - see comment above. Thanks ...

P. S.
This is during my effort to make NsCDE fully FVWM3 enabled. I have some notes about FvwmPager inconsistent behaviour that I saw and can reproduce, I will make issue reports during the day ...

ThomasAdam:
Hi @NsCDE,

Thanks!  I've attached with this message, a diff which is on top of your latest change, which fixes the `const` vs `char` confusion, as well as makes `sock_pathname` assigned once in `main()` and used where needed.  Before, you were declaring `sock_pathname` as a static global and reassigning it in different functions.  This would have been a memory leak, and also not necessary as we only need to declare it once in `main()` anyway.

Please take a look at the patch.  You can apply it on top of your work by saving the attached file and running:

`git apply /path/to/patch`

I've not tested this, so if you could please check it's doing the right thing, and then incorporate those changes in to your work, I'll take one last look and we'll get this merged in.

Thanks again!

[ontop.diff.txt](https://github.com/fvwmorg/fvwm3/files/5193786/ontop.diff.txt)

ThomasAdam:
Hi @NsCDE,

Thanks for the change.  Are you happy that it's working as it should now?

NsCDE:
> Hi @NsCDE,
> 
> Thanks! I've attached with this message, a diff which is on top of your latest change, which fixes the `const` vs `char` confusion, as well as makes `sock_pathname` assigned once in `main()` and used where needed. Before, you were declaring `sock_pathname` as a static global and reassigning it in different functions. This would have been a memory leak, and also not necessary as we only need to declare it once in `main()` anyway.
> 
> Please take a look at the patch. You can apply it on top of your work by saving the attached file and running:
> 
> `git apply /path/to/patch`
> 
> I've not tested this, so if you could please check it's doing the right thing, and then incorporate those changes in to your work, I'll take one last look and we'll get this merged in.
> 
> Thanks again!
> 
> [ontop.diff.txt](https://github.com/fvwmorg/fvwm3/files/5193786/ontop.diff.txt)

Hi Thomas,

I have applied patch and tested how things are working. It compiles fine, but after testing and reading FvwmMFL process with lsof and strace, I concluded that I have to put back `sock_pathname = set_socket_pathname();` in functions HandleTerminate and fvwm_read, because unlink gets empty string or garbage without it:
Strace:
/tmp/bla:85338 unlink("")                        = -1 ENOENT (No such file or directory)
/tmp/bla2:86157 unlink("\300\302,\1")             = -1 ENOENT (No such file or directory)

Here is the strace with set_socket_pathname call to fill sock_pathname put back: 

/tmp/bla3:86457 unlink("/tmp/fvwm_mfl.sock")      = 0
/tmp/bla4:87264 unlink("/home/tmangrow/.NsCDE/tmp/fvwm_mfl.sock") = 0

Everything now here seems to be ok, **EXCEPT** one issue: You are handling the start of FvwmMFL module internally, not from the user configuration (Init or Start conf functions). If FVWM3 configuration is used ($FVWM_USERDIR/config) to put:

`Test (EnvMatch FVWM_IS_FVWM3 1) SetEnv FVWMMFL_SOCKET $[FVWM_USERDIR]/tmp/fvwm_mfl.sock`

This will not work, because FvwmMFL is started **prior** to that. Obvious workaround is to KillModule FvwmMFL and Module FvwmMFL from the StartFunction, but this leaves unnecessary creation and closing of socket in /tmp and recreation of the process. It looks like this is in fvwm/fvwm3.c, but I'm not sure how to solve this.

NsCDE:
> Hi @NsCDE,
> 
> Thanks for the change. Are you happy that it's working as it should now?

You are really fast to reply. :-) Please see my answer above, which I wrote a minute ago.

ThomasAdam:
> Hi Thomas,

hELLO,
 
> I have applied patch and tested how things are working. It compiles fine, but after testing and reading FvwmMFL process with lsof and strace, I concluded that I have to put back `sock_pathname = set_socket_pathname();` in functions HandleTerminate and fvwm_read, because unlink gets empty string or garbage without it:

Look at the patch embedded in this message,  The `HandleTerminate()` problem is because we're declaring `sock_pathname` too late, so I've moved this up a bit.  If this is still failing for `fvwm_read` that's a different bug and one that needs looking at -- rather than just reassigning `sock_pathname = set_socket_pathname()'` each time, which in itself is a bug, since you need to `free()` the assignment first of all.

> Everything now here seems to be ok, **EXCEPT** one issue: You are handling the start of FvwmMFL module internally, not from the user configuration (Init or Start conf functions). If FVWM3 configuration is used ($FVWM_USERDIR/config) to put:
> 
> `Test (EnvMatch FVWM_IS_FVWM3 1) SetEnv FVWMMFL_SOCKET $[FVWM_USERDIR]/tmp/fvwm_mfl.sock`
> 
> This will not work, because FvwmMFL is started **prior** to that. Obvious workaround is to KillModule FvwmMFL and Module FvwmMFL from the StartFunction, but this leaves unnecessary creation and closing of socket in /tmp and recreation of the process. It looks like this is in fvwm/fvwm3.c, but I'm not sure how to solve this.

Well, I was going for always having that module running so that users didn't need to.  I've commented this out for now.  Have a look at this, and let me know how you get on.  Again, if you feel this is working correctly, commit the changes and resubmit.  If you feel that NOT starting `FvwmMFL` is the right thing to do, and making that something the user needs to do, then I'm OK with that.  Please make sure your PR includes a `git revert 4a30c4fbdfd258012dbf612f005485b2e2f2bd9f`.

Here's the patch on top of your latest changes:

```
diff --git a/fvwm/fvwm3.c b/fvwm/fvwm3.c
index 68702e0a3..b6dc82f9c 100644
--- a/fvwm/fvwm3.c
+++ b/fvwm/fvwm3.c
@@ -1396,7 +1396,6 @@ static void SetRCDefaults(void)
                { "Key Up M A MenuMoveCursor -1", "", "" },
                { "Key Down M A MenuMoveCursor 1", "", "" },
                { "Mouse 1 MI A MenuSelectItem", "", "" },
-               { "Module FvwmMFL", "", "" },
                /* don't add anything below */
                { RC_DEFAULTS_COMPLETE, "", "" },
                { "Read "FVWM_DATADIR"/ConfigFvwmDefaults", "", "" },
diff --git a/modules/FvwmMFL/FvwmMFL.c b/modules/FvwmMFL/FvwmMFL.c
index 722421af7..52fd210c9 100644
--- a/modules/FvwmMFL/FvwmMFL.c
+++ b/modules/FvwmMFL/FvwmMFL.c
@@ -682,6 +682,8 @@ int main(int argc, char **argv)
                return (1);
        }
 
+       sock_pathname = set_socket_pathname();
+
        /* Create new event base */
        if ((base = event_base_new()) == NULL) {
                fprintf(stderr, "Couldn't start libevent\n");
@@ -689,8 +691,6 @@ int main(int argc, char **argv)
        }
        setup_signal_handlers(base);
 
-       sock_pathname = set_socket_pathname();
-
        memset(&sin, 0, sizeof(sin));
        sin.sun_family = AF_LOCAL;
        strcpy(sin.sun_path, sock_pathname);
```

NsCDE:
Thomas:

> Look at the patch embedded in this message. The `HandleTerminate()` problem is because we're declaring `sock_pathname` too late, so I've moved this up a bit. If this is still failing for `fvwm_read` that's a different bug and one that needs looking at -- rather than just reassigning `sock_pathname = set_socket_pathname()` each time, which in itself is a bug, since you need to `free()` the assignment first of all.

I have reverted this and removed calls from `HandleTerminate()` and `fvwm_read()`. Removed call to FvwmMFL in fvwm/fvwm3.c and moved last remaining `sock_pathname = set_socket_pathname()` up where you pointed it in this simple patch.

Outcome:

This unlink() problem still persists. Socket remains in directory and on the second call to "Module FvwmMFL" this fails with "Couldn't create listener: Address already in use".

P. S.
As expected, FvwmMFL is not started if not called from the configuration, this part is ok, and environment is picked up when StartFunction with modules is executing from the config.

NsCDE:
Replying myself:

> Outcome:
> 
> This unlink() problem still persists. Socket remains in directory and on the second call to "Module FvwmMFL" this fails with "Couldn't create listener: Address already in use".
> 
> P. S.
> As expected, FvwmMFL is not started if not called from the configuration, this part is ok, and environment is picked up when StartFunction with modules is executing from the config.

Turns out this is not the problem of HandleTerminate, but unlink from the end of `main()`. The cause is: `free(sock_pathname);` right next after `strcpy(sin.sun_path, sock_pathname);` on line 692. I'm not sure where this
free should go. Probably before return in main and AFTER unlink. :-) Sorry for my poor C ...

ThomasAdam:
> EDIT ... if anywhere this free() should go. Program terminates here, so it is freed if I'm correct.

I'm tired.  Forgive me.  No need at all to `free()` here -- the variable needs to persist!  Please remove that.

NsCDE:
Hi.

Here is then final PR proposal. And have some rest. Tomorrow I will try to make some torture with WindowName patch for FvwmButtons. :-)

ThomasAdam:
Hi @NsCDE,

So.  I've squashed and rebased your commits together, and came up with:

https://github.com/fvwmorg/fvwm3/pull/217

I've retained your original authorship, so things are rightly accredited to yourself.  Please have a look at that.  If you're happy, Please close this PR, mention me on the one above and I'll merge that, and everything will be hunky-dory as they say.

Kindly,
Thomas

NsCDE:
Thanks Thomas, and there is no necessity accreditation if something is rebased technically, I'm glad to help with FVWM.

Closing this, and going to mention you on the #217 whatever this means. :-)

PR #217, opened 2020-09-09
ta/gh 214

ThomasAdam:
- Revert "FvwmMFL: start by default"
- Respect environment variable for MFL socket path


NsCDE:
Hi @ThomasAdam 

This looks nice now. Maybe in the future some unlink before opening pathname should exist, because program cannot cleanup on power outage or sigkill. For now, Exec exec rm -f <pathname> prior to "Module FvwmMFL" should do the trick.

PR #219, opened 2020-09-09
Add WindowName option to FvwmButtons

NsCDE:
Hi,

Today I have adapted "WindowName" FvwmButtons patch for FVWM3, together with man page entry.
I have been using it with FVWM2 for 1-2 years already.

This adds functionality for subpanels which intentionally have Title (like CDE Front Panel
and it's subpanels as you can guess) with non-ASCCI characters and spaces, which cannot be
accomplished with FvwmButtons module alias.

### DIFF
--- a/modules/FvwmButtons/FvwmButtons.1.in
+++ b/modules/FvwmButtons/FvwmButtons.1.in
@@ -181,6 +181,16 @@ double quotes) for a transparent background.
 Specifies the number of rows of buttons to be created. The default
 is 2 rows.
 
+.IP "*FvwmButtons: WindowName \fIname\fP"
+If FvwmButtons has a titlebar enabled with Title style, (for example,
+some transient subpanel), this option can set it's Window name and
+Icon name to a string provided with this parameter. If omitted, default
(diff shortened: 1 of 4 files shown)


ThomasAdam:
Hi @NsCDE,

This looks good.  I'm note sure `char windowname[128];` is necessarily long enough, but I suppose it will do for now.   One suggestion:

```
diff --git a/modules/FvwmButtons/parse.c b/modules/FvwmButtons/parse.c
index 394399518..dac8f2b44 100644
--- a/modules/FvwmButtons/parse.c
+++ b/modules/FvwmButtons/parse.c
@@ -1946,10 +1946,7 @@ static void ParseConfigLine(button_info **ubb, char *s)
                        ub->c->flags.b_WindowName = 0;
                }
 #ifdef DEBUG_PARSER
-               if (windowname == 0)
-               {
                        fprintf(stderr, "%s\n", windowname);
-               }
 #endif
                break;
        }
```

NsCDE:
Hi,

Removed unnecessary if (windowname ... and resubmitted.

PR #221, opened 2020-09-10
Add support for triangle indicator in/out sub options

NsCDE:
Hi Thomas. This is my long existing patch 2/3 it messes with the same files as window name patch, so I have split it and sent as separate patch.

What does it do: it adds ability for subpanel launcher triangle indicator to be drawn in 3D sunken in, not just raised out. This means that indicator has new argument - after positive integer, there can be keyword "in" or "out". Absence of this argument means "out" which is visual and configuration parsing backward compatible behaviour.

A small illustration attached: right is default raised triangle, left is sunken in new option.
FvwmButtons-Patch-in-out-detail

### DIFF
--- a/modules/FvwmButtons/parse.c
+++ b/modules/FvwmButtons/parse.c
@@ -396,7 +396,8 @@ static void ParsePanel(
 	char **ss, unsigned int *flags, unsigned int *mask, char *direction,
 	int *steps, int *delay, panel_flags_type *panel_flags,
 	int *indicator_size, int *rela_x, int *rela_y, char *position,
-	char *context)
+	char *context, int *indicator_sunkraise, char *indicator_in_out)
+
 {
 	char *swallowopts[] =
 	{
@@ -424,6 +425,9 @@ static void ParsePanel(
 
 	char *t, *s = *ss;
 	int n;
+	char *instr = "in";
+	char *outstr = "out";
+	char *clear_comma;
 
 	while (*s && *s != ')')
 	{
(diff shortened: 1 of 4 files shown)


modules/FvwmButtons/parse.c
```diff
@@ -515,7 +519,38 @@ static void ParsePanel(
...
 			*indicator_size = 0;
-			sscanf(s, "%d%n", indicator_size, &n);
+			sscanf(s, "%d %s%n", indicator_size, indicator_in_out, &n);
+
+			/* Remove comma from %s of sscanf */
+			clear_comma = strchr(indicator_in_out, ',');
+			if (clear_comma != NULL)
+			{
+				*clear_comma = '\0';
+			}
```
ThomasAdam, line 529: Maybe:

```suggestion
			indicator_in_out[strcspn(indicator_in_out, ",")] = '\0';
```

ThomasAdam (commented):
Hi @NsCDE,

Looks good.

modules/FvwmButtons/parse.c
```diff
@@ -515,7 +519,38 @@ static void ParsePanel(
 			n = 0;
 			(*panel_flags).panel_indicator = 1;
 			*indicator_size = 0;
-			sscanf(s, "%d%n", indicator_size, &n);
+			sscanf(s, "%d %s%n", indicator_size, indicator_in_out, &n);
```
ThomasAdam, line 522: We might want to change this string to put size limits?  `"%d %s<NUM>%n"` -- currently `%s` is unbounded, so could crash with large input.

NsCDE:
Hi,

I have implemented your first suggestion.

But I cannot find the second one in the code. I'm bit disoriented now. Also for size limits ... how many to put for this ... please resolve if you can. :-)

ThomasAdam (approved):

ThomasAdam:
Hi @NsCDE,

Not a problem.  We'll monitor this for now. Let's see how we get on!  Merged now.

PR #222, opened 2020-09-11
Support for setting default mouse cursor on all FvwmScript widgets

NsCDE:
Hi Thomas.

This is patch 3/3. I had to rework it as configurable option, and not just to re-hardcode - to be acceptable for wider public.

Why:
Most of the FvwmScript widgets (push buttons, menus, checkboxes ...) are changing default cursor into XC_hand2. This is interesting but weird choice because no any other widget in my knowledge is using cursor symbol for hyperlinks, URLs, or pulling something (hand2) on objects/widgets of such nature. It looks weird.

However:
This default lasts long years, and there may be users who likes XC_hand2 on widgets or got used to this. Because of that, this patch is non-intrusive.

How:
It adds new FvwmScript configuration option "*FvwmScript DefaultCursor" which argument is cursor name from X11/cursorfont.h without "XC_" prefix - just like in CursorStyle FVWM option (it uses the same API from libs/Cursor.c).

Behaviour:
* if DefaultCursor option is present and valid, it changes cursor on widgets (for example to left_ptr)
* if DefaultCursor option is present and invalid, it is discarded and old behaviour takes a place
* if DefaultCursor option is not present, old behaviour takes a place

P. S.
I hope I was careful not to leave leaks and corpses behind it! :-)

### DIFF
--- a/modules/FvwmScript/FvwmScript.c
+++ b/modules/FvwmScript/FvwmScript.c
@@ -39,6 +39,7 @@
 #include "libs/Parse.h"
 #include "libs/Strings.h"
 #include "libs/XError.h"
+#include "libs/Cursor.h"
 
 #ifdef MEMDEBUG                 /* For debugging */
 #include <unchecked.h>
@@ -97,6 +98,7 @@ Binding *BindingsList;
 extern void InitCom(void);
 static void TryToFind(char *filename);
 
+char *default_cursor;
 
 /* Exit procedure - called whenever we call exit(), or when main() ends */
 static void
@@ -281,6 +283,11 @@ void ParseOptions(void)
 	CopyStringWithQuotes(&x11base->font,&tline[22]);
       else if (strncasecmp(tline,"Colorset",8) == 0)
 	LoadColorset(&tline[8]);
+      else if (strncasecmp(tline,"*FvwmScriptDefaultCursor",24) == 0)
+      {
+	CopyString(&x11base->cursor,&tline[24]);
+	default_cursor = x11base->cursor;
+      }
     }
 
     GetConfigLine(fd,&tline);
@@ -439,6 +446,7 @@ void OpenWindow (void)
   unsigned long mask;
   XClassHint classHints;
   XSetWindowAttributes Attr;
+  int xcnum;
 
   /* Allocation for colors */
   if (x11base->colorset >= 0) {
@@ -458,6 +466,16 @@ void OpenWindow (void)
   Attr.background_pixel = x11base->TabColor[back];
   Attr.border_pixel = 0;
   Attr.colormap = Pcmap;
+
+  xcnum = fvwmCursorNameToIndex(default_cursor);
+  if (xcnum != -1) {
+    Attr.cursor = XCreateFontCursor(dpy, xcnum);
+    mask |= CWCursor;
+  } else {
+    /* Bad name or value, clear the config slot. */
+    x11base->cursor = NULL;
+  }
+  free(default_cursor);
 
   x11base->win=XCreateWindow(
     dpy, x11base->root, x11base->size.x,x11base->size.y, x11base->size.width,
--- a/modules/FvwmScript/FvwmScript.1.in
+++ b/modules/FvwmScript/FvwmScript.1.in
@@ -85,6 +85,11 @@ command or in the script with the ShadowColor command, grey55 is used.
 .IP "*FvwmScript: DefaultColorset \fIcolorset\fP"
 Tells the module to use colorset \fIcolorset\fP as the default colorset.
 
+.IP "*FvwmScript: DefaultCursor \fIcursor\fP"
+Tells the module to use X11 cursor \fIcursor\fP as the default for all
+widgets. Cursor names are those defined in the include file X11/cursorfont.h
+but without XC_ prefix.
+
 .SH ANATOMY OF A SCRIPT
 FvwmScript uses a particular programming language. A script is composed of
 five parts. Heading contains general characteristics of the window and
--- a/modules/FvwmScript/Widgets/CheckBox.c
+++ b/modules/FvwmScript/Widgets/CheckBox.c
@@ -44,8 +44,10 @@ void InitCheckBox(struct XObj *xobj)
   mask = 0;
   Attr.background_pixel = xobj->TabColor[back];
   mask |= CWBackPixel;
-  Attr.cursor = XCreateFontCursor(dpy,XC_hand2);
-  mask |= CWCursor;  /* Cursor for the window */
+  if (!x11base->cursor) {
+    Attr.cursor = XCreateFontCursor(dpy,XC_hand2);
+    mask |= CWCursor;  /* Window cursor */
+  }
 
   xobj->win = XCreateWindow(dpy, *xobj->ParentWin,
 		xobj->x, xobj->y, xobj->width, xobj->height,0,
--- a/modules/FvwmScript/Widgets/HScrollBar.c
+++ b/modules/FvwmScript/Widgets/HScrollBar.c
@@ -93,8 +93,10 @@ void InitHScrollBar(struct XObj *xobj)
   mask = 0;
   Attr.background_pixel = xobj->TabColor[back];
   mask |= CWBackPixel;
-  Attr.cursor = XCreateFontCursor(dpy,XC_hand2);
-  mask |= CWCursor;             /* cursor for window */
+  if (!x11base->cursor) {
+    Attr.cursor = XCreateFontCursor(dpy,XC_hand2);
+    mask |= CWCursor;  /* Window cursor */
+  }
 
   xobj->win = XCreateWindow(dpy, *xobj->ParentWin,
 			    xobj->x, xobj->y, xobj->width, xobj->height, 0,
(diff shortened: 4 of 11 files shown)


modules/FvwmScript/FvwmScript.c
```diff
@@ -281,6 +283,11 @@ void ParseOptions(void)
 	CopyStringWithQuotes(&x11base->font,&tline[22]);
       else if (strncasecmp(tline,"Colorset",8) == 0)
 	LoadColorset(&tline[8]);
+      else if (strncasecmp(tline,"*FvwmScriptDefaultCursor",24) == 0)
```
ThomasAdam, line 286: Do you really need the leading `*` here?

Also, can we not just call it `DefaultCursor`?

ThomasAdam (changes requested):
Hi @NsCDE,

Yep -- all looks good.  Before I even looked at the change, the diffstat looked quite large for configuring cursors, but then I can see there's duplication for each widget.

One tiny query, otherwise the rest looks good.

modules/FvwmScript/FvwmScript.c
NsCDE: > Hi @NsCDE,

Hi Thomas,

Sorry for a delay. It went to sleep. :-)

I will answer about both issues you highlighted here:

> Yep -- all looks good. Before I even looked at the change, the diffstat looked quite large for configuring cursors, but then I can see there's duplication for each widget.

It was like that before, and wanting to be non-intrusive, I have simply armored this hard coding of cursor with my new condition in all that places.

> One tiny query, otherwise the rest looks good.
> Do you really need the leading * here?
> Also, can we not just call it DefaultCursor?

This part of the configuration parsing code is not really clear to me. If you look above line 286 in FvwmScript.c you will see that I have simply mimicked the way other global FvwmScript options are defined. Looks like this "*FvwmScript<bla>" part of the string is someware somehow split into "*FvwmScript: " and "DefaultCursor" afterwards, because this is exactly what I have put in configuration while testing this:

`*FvwmScript: DefaultCursor left_ptr`

P. S.
Soon, when I catch some time there there will be TextField colorset unwanted inheritance of foreground color bug report in FvwmScript. I will try to figure it out and fix myself, but it doesn't look really readable, so maybe it will be an issue. This is bug also in 2.X.

danespen:
Gee, the reply addresses from github are a mess.
I deleted all of them except the reply.gihub.com address.
Don't know if that will work.


Hegel3DReloaded <notifications@github.com> writes:

> @NsCDE commented on this pull request.
>
> -----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
>
> In modules/FvwmScript/FvwmScript.c:
>
>> @@ -281,6 +283,11 @@ void ParseOptions(void)
>  	CopyStringWithQuotes(&x11base->font,&tline[22]); else if
>        (strncasecmp(tline,"Colorset",8) == 0)
>  	LoadColorset(&tline[8]); + else if
> (strncasecmp(tline,"*FvwmScriptDefaultCursor",24) == 0)
>
>  Hi @NsCDE,
>
> Hi Thomas,
>
> Sorry for a delay. It went to sleep. :-)
>
> I will answer about both issues you highlighted here:
>
>  Yep -- all looks good. Before I even looked at the change, the
>  diffstat looked quite large for configuring cursors, but then I can
>  see there's duplication for each widget.
>
> It was like that before, and wanting to be non-intrusive, I have
> simply armored this hard coding of cursor with my new condition in all
> that places.
>
>  One tiny query, otherwise the rest looks good.  Do you really need
>  the leading * here?  Also, can we not just call it DefaultCursor?
>
> This part of the configuration parsing code is not really clear to
> me. If you look above line 286 in FvwmScript.c you will see that I
> have simply mimicked the way other global FvwmScript options are
> defined. Looks like this "*FvwmScript" part of the string is someware
> somehow split into "*FvwmScript: " and "DefaultCursor" afterwards,
> because this is exactly what I have put in configuration while testing
> this:
>
> *FvwmScript: DefaultCursor left_ptr

What FvwmScript is doing is keeping it's "global" configuration
in the config for "*FvwmScript".  Like most of the modules you can
invoke FvwmScript using alternate names.  Even when invoked with the
alternate name, the *FvwmScript options for selected options still apply
because of the logic just extended.

This is a little different than FvwmForm that looks for it's defaults in
*FvwmFormDefault and forces a read of it's config from .FvwmForm:

  SendText(Channel,"read .FvwmForm Quiet",0); /* read default config */

So, what FvwmScript is doing is okay except a user might be surprised
when he designs a FvwmScript using the default name then another using
an alternate name and sees stuff from the default name leaking into
the alternate name.


-- 
Dan Espen

ThomasAdam:
> What FvwmScript is doing is keeping it's "global" configuration in the config for "*FvwmScript". Like most of the modules you can invoke FvwmScript using alternate names. Even when invoked with the alternate name, the *FvwmScript options for selected options still apply because of the logic just extended. This is a little different than FvwmForm that looks for it's defaults in *FvwmFormDefault and forces a read of it's config from .FvwmForm: SendText(Channel,"read .FvwmForm Quiet",0); /* read default config */ So, what FvwmScript is doing is okay except a user might be surprised when he designs a FvwmScript using the default name then another using an alternate name and sees stuff from the default name leaking into the alternate name.

Indeed.  I will probably force all modules in `Fvwm3` to adopt something similar.  I favour making `*Fvwm<ModuleName>` a default global for all `Fvwm<ModuleName>` instances.  I think there's plenty of modules where this would work -- but would obviously break existing configs.

Anyhow, in terms of that suggestion, @NsCDE -- the module name is stripped from the config lines sent down from `Fvwm`.  That's why you sometimes see older module config lines like this:

```
*FvwmButtonsPadding 1
*FvwmButtonsBoxSize 3
```

The "FvwmButtons" part was stripped, leaving the config option and the value.

This changed though to be:

```
*FvwmButtons: Padding x
```
From having the briefest of glances at `FvwmScript` before heading to the nearby village of Ilkley, I suggest you look at making the cursor changes in the lex config scanner -- `script.y`.

danespen:
For FvwmForm, I chose FVwmFormDefault so that a user wouldn't
accidentally design a form with that name and then be surprised at some
odd behavior.  If I remember right, the man page warns not to use
any name starting with "Fvwm" to avoid clashes.

I think a lot of users default behavior is to just use the module name
for their config instead of setting up an alias.  For many modules this
is just natural.  For example, you only need one FvwmAnimate.

Thomas Adam <notifications@github.com> writes:

>  What FvwmScript is doing is keeping it's "global" configuration in
>  the config for "*FvwmScript". Like most of the modules you can invoke
>  FvwmScript using alternate names. Even when invoked with the
>  alternate name, the *FvwmScript options for selected options still
>  apply because of the logic just extended. This is a little different
>  than FvwmForm that looks for it's defaults in FvwmFormDefault and
>  forces a read of it's config from .FvwmForm: SendText(Channel,"read
>  .FvwmForm Quiet",0); / read default config */ So, what FvwmScript is
>  doing is okay except a user might be surprised when he designs a
>  FvwmScript using the default name then another using an alternate
>  name and sees stuff from the default name leaking into the alternate
>  name.
>
> Indeed. I will probably force all modules in Fvwm3 to adopt something
> similar. I favour making *Fvwm<ModuleName> a default global for all
> Fvwm<ModuleName> instances. I think there's plenty of modules where
> this would work -- but would obviously break existing configs.
>
> Anyhow, in terms of that suggestion, @NsCDE -- the module name is
> stripped from the config lines sent down from Fvwm. That's why you
> sometimes see older module config lines like this:
>
> *FvwmButtonsPadding 1 *FvwmButtonsBoxSize 3
>
> The "FvwmButtons" part was stripped, leaving the config option and the
> value.
>
> This changed though to be:
>
> *FvwmButtons: Padding x
>
> From having the briefest of glances at FvwmScript before heading to
> the nearby village of Ilkley, I suggest you look at making the cursor
> changes in the lex config scanner -- script.y.
>
> — You are receiving this because you commented.  Reply to this email
> directly, view it on GitHub, or unsubscribe.
>

-- 
Dan Espen

NsCDE:
Hi Thomas & Dan,

Thank you both for this explanation. Yes, I can recall some very early examples with "*Fvwm" on the beginning. Luckily, it is very unusual to alias FvwmScript. FvwmButtons, FvwmPager and FvwmIconMan are far more common candidates in a wild.

@ThomasAdam If I understood this correctly ... I'm a bit confused with config scanner, not really familiar with lex. :-( I can try, but this means new option will then appear as global header directive in FvwmScript script code file (like WindowSize) and not in FVWM configuration. This will introduce backward incompatible change, because in FvwmScript body, we mostly don't have FVWM flexibility to "ifdef" with EnvMatch for a smooth deployment across various versions.

How to go further now with this? Please advice ...

ThomasAdam:
Hi @NsCDE,

Yes I know it will. I suppose there's a case for per widget of option here, but then again the cursors have always been global.

If someone else cares enough to rework this then they can.

NsCDE:
Hi @ThomasAdam 

Whew ... I almost got panic attack while trying to read script.y. :-)

As of cursors generally as global - I agree and will like to add: if we look at various graphical widgets, apart from more and more rare busy cursor this days, seldom will something change cursor shape for no good reason other then pulling internal borders. Programs such as Gimp or Inkscape are totally another story, where such cursor changes are totally justified. I was always wandering why XC_hand2 (used in browsers for hyperlinks) was choosen for ordinary push buttons, check boxes and menus anyway.

PR #239, opened 2020-10-02
Add variable/parameter which indicates the state of the FVWM3 logging

NsCDE:
Hi @ThomasAdam 

I realized that there is no way to check if logging and debugging with SIGUSR2 is currently enabled or disabled. This may lead to the situation that something opens the log, crash and not closes it, then "opens" it again, but actually closes it.
This is my proposal for new internal varible/parameter `$[debuglog.state]` with value 0 or 1 for checking this state.

P. S.
In libs/log.h I already found `log_get_level()` declared but unused in log.c when I wanted to declare it. So I'm probably on the trace of some forgotten idea. :-)

### DIFF
--- a/doc/fvwm/expansion.xml
+++ b/doc/fvwm/expansion.xml
@@ -518,6 +518,15 @@ command can be used:</para>
     </listitem>
   </varlistentry>
   <varlistentry>
+    <term>$[debuglog.state]</term>
+    <listitem>
+      <para>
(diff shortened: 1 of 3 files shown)


ThomasAdam (changes requested):
Hi @NsCDE,

This all looks good.  Thanks.

ThomasAdam (approved):

Issue #246, opened 2020-10-10
Man page `fvwm3.1` not built by default


djwf:
When making and installing fvwm3 from the git repository, using the following commands, the file `fvwm3.1` is not built or installed.

```sh
./autogen.sh
./configure
make
make install
```

This is unexpected given the first sentence of the second paragraph in `doc/README`: "By default, only the man page is generated."

If I add `--enable-mandoc` to the configure command, and run the commands below, then the `fvwm3.1` file is created.

```sh
./autogen.sh
./configure --enable-mandoc
make
make install
```

This behavior may be intentional, and I figured I'd check before submitting a PR to fix either the `README` *or* the `Makefile.am`.

So: is the behavior intentional?

ThomasAdam:
Hi @djwf 

Hehe! Thanks for your eagie eyes! It's deliberate, and the README is wrong. Please submit a PR. Documentation is now opt-in.

Kindly,
Thomas

djwf:
Got it - so should all the man pages be omitted by default? Or just the main page? I ask to figure out what should be in the PR.

ThomasAdam:
Hi @djwf 

It's technically just the main fvwm3 page as that's what is built via the snake which is docbook.

I wouldn't draw a distinction though, I'd rather the PR just removed the assumption that manages are built for free.

djwf:
Got it, PR will be inbound shortly!

PR #247, opened 2020-10-10
Revise doc/README - docs not built by default

djwf:
While here I reordered the mandoc/htmldoc commands so the next part (about index.html) makes more sense.

Oh, and this deals with #246.

### DIFF
--- a/doc/README
+++ b/doc/README
@@ -1,19 +1,20 @@
 # Author: Scott Smedley
 
-This directory contains fvwm XML documentation. It is configured
-to output an fvwm.1 man page & a plethora of HTML files - all from
-the same (XML) source.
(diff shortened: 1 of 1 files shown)


ThomasAdam:
Hi @djwf 

Thanks. This looks good to me.

Can you expand on the commit message a little tough, and add:

`Fixes #xxx`

To the end of the commit message so that the issue you raised is tracked? This will really help me when the changelog is generated for the next release. Make sure you replace xxx with the correct number!

djwf:
Yep - clarified purpose of change and added Fixes note.

ThomasAdam:
Thanks @djwf - merged!

PR #263, opened 2020-10-18
Fix module synchronization in goto_desk

### LINKED ISSUE #262: FvwmPager and FvwmIconMan do not update after GotoDesk
After upgrading to Fvwm3 1.0.1 from 1.0.0, FvwmPager and FvwmIconMan do not update after switching to other desk with GotoDesk. They keep displaying the old window configuration instead. Clicking on FvwmPager virtual desk resolves the issue.

The issue happens with the default configuration. For example:

- On Desk 0, open any window.
- Change to Desk 1 by clicking on FvwmPager desk indicator.
- Observe how Desk 0 window configuration stays on the Pager window.
- Click on the FvwmPager miniature desk window (eg. change to another page) and observe how window configuration is updated.

I bisected this fault to commit e557dd3d5f83caab349d52fe9afee8375514798e
By commenting out if-statement on virtual.c:1537 the problem is resolved.
The bug occurs only when DisplayConfiguration is global.

This may be related to #223.

phintsan:
Change monitor looping in goto_desk so that modules on the current
monitor will also get broadcasted packets.

Fixes #262.

### DIFF
--- a/fvwm/virtual.c
+++ b/fvwm/virtual.c
@@ -1534,11 +1534,10 @@ void goto_desk(int desk, struct monitor *m)
 		}
 
 		TAILQ_FOREACH(m2, &monitor_q, entry) {
-			if (m == m2)
-				continue;
-
-			m2->Desktops = m->Desktops;
-			m2->virtual_scr.CurrentDesk = m->virtual_scr.CurrentDesk;
+			if (m != m2) {
+				m2->Desktops = m->Desktops;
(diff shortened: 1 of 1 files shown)


ThomasAdam:
Hi @phintsan,

Thanks for this. I think the commit message is fine. You're right to have that if statement... I generally prefer not to self-assign when I don't have to.

PR #277, opened 2020-10-26
Remove SAFEFREE macro

tekknolagi:
Replace all uses with free().

It looks like most uses in this file immediately set the field being freed.

### DIFF
--- a/fvwm/style.c
+++ b/fvwm/style.c
@@ -38,8 +38,6 @@
 
 /* ---------------------------- local definitions -------------------------- */
 
-#define SAFEFREE( p )  {if (p) {free(p);(p)=NULL;}}
-
 /* ---------------------------- local macros ------------------------------- */
 
 /* ---------------------------- imports ------------------------------------ */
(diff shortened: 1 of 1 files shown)


ThomasAdam:
Hi @tekknolagi 

Thank you for this! Have you noticed any breakages with this change? That is, crashes because via SAFEFREE we were setting the free()d variable to NULL, and now we're not?

I'm guessing not, but it wouldn't surprise me.

tekknolagi:
I have not attempted to run FVWM3 with this change, no.

I can go back and change the instances that do not immediately modify fields after SAFEFREE to manually set them to NULL, if you like.

ThomasAdam:
Hi @tekknolagi 

No, it's OK. I was curious. It's probably easier to merge this and fix up anything relying on NULL-after-free.

tekknolagi:
Cool! Thanks :) If you would not mind tagging this PR with `hacktoberfest-accepted`, that would be appreciated!

ThomasAdam:
Hi @tekknolagi 

Oh I see. I did wonder what the motivation for the change was, given you said you hadn't run fvwm3.

Hmm.

tekknolagi:
Well the motivation was not that, surprisingly :P I saw it come up on my feed, saw there was a todo, and figured "how hard could this be?"

Then, after the fact, remembered that. But also really, no need

tekknolagi:
(todo from https://github.com/fvwmorg/fvwm3/blob/master/dev-docs/TODO.md )

PR #281, opened 2020-10-31
FvwmForm: remove useless setitimer call

phileimer:
As the initial setitimer is configured with a periodic restart, the call in TimerHandler is useless.

### DIFF
--- a/modules/FvwmForm/FvwmForm.c
+++ b/modules/FvwmForm/FvwmForm.c
@@ -100,7 +100,7 @@ int colorset = -1;
 int itemcolorset = 0;
 
 /* global not exported */
-const struct itimerval itv_100ms = { {0L, 100000L}, {0L, 100000L} };
+const struct itimerval itv_100ms = { {0L, 100000L}, {0L, 100000L} }; /* interval = 100ms, value = 100ms */
 
 /* prototypes */
 static void RedrawSeparator(Item *item);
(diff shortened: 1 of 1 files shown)


modules/FvwmForm/FvwmForm.c
```diff
@@ -2607,9 +2607,9 @@ static void MainLoop(void)
     /*XFlush(dpy);*/
     if (fvwmSelect(fd_width, &fds, NULL, NULL, NULL) > 0) {
       if (FD_ISSET(Channel[1], &fds))
-	ReadFvwm();
+        ReadFvwm();
       if (FD_ISSET(fd_x, &fds))
-	ReadXServer();
+        ReadXServer();
```
ThomasAdam, line 2612: This chunk is part of an unrelated change.  Please elide it from the commit.

modules/FvwmForm/FvwmForm.c
```diff
@@ -100,7 +100,7 @@ int colorset = -1;
 int itemcolorset = 0;
 
 /* global not exported */
-const struct itimerval itv_100ms = { {0L, 100000L}, {0L, 100000L} };
+const struct itimerval itv_100ms = { {0L, 100000L}, {0L, 100000L} }; /* interval = 100ms, value = 100ms */
```
ThomasAdam, line 103: This comment isn't part of timer removal, so should be elided.  It's probably not even worth including as separate patch.

ThomasAdam (changes requested):
Hi @phileimer 

Thanks -- looks good to me.  Please see a few comments regarding unrelated changes.

PR #282, opened 2020-10-31
FvwmForm: use posix timer

phileimer:
The actual FvwmForm timer code is based on setitimer and signal management.
The setitimer man page lists the following bug:
"The generation and delivery of a signal are distinct, and only one instance of each of the signals listed above may be pending for a process. Under very heavy loading, an ITIMER_REAL timer may expire before the signal from a previous expiration has been delivered. The second signal in such an event will be lost."

I use FvwmForm to display the desktop name, for 400ms, each time desktop changes.
By changing desktop very often I - very rarely - manage to have a FvwmForm hang, lost in an infinite loop, with a 99% cpu usage. I then have to kill it with -KILL.
I debugged to find there was a missing TimerHandler call
So, I thought this was due to a signal loss as mentioned above. 

I tried to modify the code to solve that, and also because setitimer is considered deprecated.

Unfortunately, I still have the FvwmForm hang, with the same rare occurrence.

I let @ThomasAdam decide whether this PR is of interest.

### DIFF
--- a/modules/FvwmForm/FvwmForm.c
+++ b/modules/FvwmForm/FvwmForm.c
@@ -100,7 +100,8 @@ int colorset = -1;
 int itemcolorset = 0;
 
 /* global not exported */
-const struct itimerval itv_100ms = { {0L, 100000L}, {0L, 100000L} }; /* interval = 100ms, value = 100ms */
+const struct itimerspec its_100ms = { {0L, 100000000L}, {0L, 100000000L} }; /* interval = 100ms, value = 100ms */
+static timer_t timerid;
 
 /* prototypes */
 static void RedrawSeparator(Item *item);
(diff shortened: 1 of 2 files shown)


ThomasAdam (commented):
Hi @phileimer 

Wasn't this commit submitted as a separate PR?

ThomasAdam (commented):
Hi @phileimer 

I think this patch is on the right lines.  However, I think it would be much more appropriate to make more use of libevent, seeing as this is already a dependency on Fvwm3, and would abstract away all of the existing compact code for signal handling.

phileimer:
Yes, the first commit is another PR. My mistake. Sorry.

ThomasAdam:
Right -- then this can be rebased once it's through code review.  No problem!

phileimer:
Thank you for the libevent idea. When I can, I'll try it to check whether I can remove the hang bug.

phileimer:
I close this, waiting the work based on libevent.

PR #284, opened 2020-11-01
FvwmForm: remove useless setitimer call

phileimer:
As the initial setitimer is configured with a periodic restart, the call in TimerHandler is useless.


ThomasAdam:
Hi @phileimer 

Fantastic. Much better. Will merge once the CI checks pass.

Thanks again.

Thomas

PR #289, opened 2020-11-21
FvwmForm: use libevent for timer instead of signal

phileimer:
With the current signal based timer, FvwmForm sometimes falls into an infinite loop.
Using libevent is a tentative to avoid this hang.


ThomasAdam (approved):
Hi,

Thanks for revisiting this.  It looks good, and finally removes a lot of old/broken signal-handling code.

We should do something similar across all the remaining modules.

PR #293, opened 2020-11-29
clean up the headers

mikeandmore:
Make all headers self-contained so that we can use many clang-based tools to analyze the code.


ThomasAdam:
Hi @mikeandmore,

Looks good.  As mentioned on IRC, worth considering the include guards using a `FVWM_` prefix.

I'm still suggesting we should handle inclusion of X11/Xlib header files better.  They're all over the place with this change at the moment.

I would also recommend you run something like `include-what-you-use` or [deheader](http://www.catb.org/~esr/deheader/) over your change.  I use `deheader` like this:

```
make clean
deheader -m "make -j $(nproc) CFLAGS='-Werror -Wfatal-errors'"
```

mikeandmore:
let me try include-what-you-use. Now I can run clang based tools on the headers.

mikeandmore:
include-what-you-use suggests to add more X11 headers instead of removing them.

Do you think it's a good idea to add a `common_x11.h` in `libs/`?

ThomasAdam:
Hi @mikeandmore,

> include-what-you-use suggests to add more X11 headers instead of removing them.
> 
> Do you think it's a good idea to add a `common_x11.h` in `libs/`?

Absolutely.  This is what I was getting at on IRC -- we're including them sporadically throughout the code base right now, but it should be referenced via something like `libs/fvwm_x11.h`.

Cheers!

Issue #164, opened 2020-07-10
Safe sprintf call


klebertarcisio:
Hi everyone,

I would like to know what do you think about these sprintf calls:

https://github.com/fvwmorg/fvwm3/blob/48dd50909b97170d517aa02c735cc86881564cb6/modules/FvwmIdent/FvwmIdent.c#L1332-L1335

In this case, x and y are integer variables, but max_aspect is a char of just 11 bytes.
See this toy example:
**sprintf (mymin_aspect, "% d % d", 111111, 111111);**
In this case, we have an overflow. Should the **mymin_aspect** variable be **24** bytes?

https://github.com/fvwmorg/fvwm3/blob/48dd50909b97170d517aa02c735cc86881564cb6/libs/Colorset.c#L111-L120

In this case, **csetbuf** is a char of just **256** bytes, but the operation requires **270** bytes.

https://github.com/fvwmorg/fvwm3/blob/48dd50909b97170d517aa02c735cc86881564cb6/fvwm/icccm2.c#L56-L56

In this case, **wm_sx** is a char of just **20** bytes, but the operation requires **25** bytes.

Regards

ThomasAdam:
Hi @klebertarcisio,

Yes, the compiler is starting to spit these warnings out now, as is the static analysis tool we're using.  See:

https://github.com/fvwmorg/fvwm3/issues/107

I'm all for fixing these, however, what I'd rather see are things like this changed as a result of bug-fixes in that area of the code, rather than necessarily going through all of these things piecemeal.

A lot of what you're identifying is code which has been there for many years, and although compilers are now getting smarter, the code itself isn't causing any problems.  We will fix these, but I'd much rather see some of the issues blocking a 1.0 release looked at first:  https://github.com/fvwmorg/fvwm3/projects/1?card_filter_query=milestone%3A1.0

ThomasAdam:
Hi @klebertarcisio,

Now that things have started to settle -- can you revisit this?

klebertarcisio:
Hi @ThomasAdam, yes, I can do it!

ThomasAdam:
Hi @klebertarcisio 

I hope you're well.  Just checking to see if you think you'll be able to look at this issue soon?  If not, I'll close this and move this in to the TODO list.

No pressure -- just trying to do some project maintenance.

klebertarcisio:
[...]

PR #306, opened 2020-12-01
Increase memory space to avoid overflow

klebertarcisio:
This pull request is concerned with avoiding overflows in this project. For this, I increased the memory space of some variables to ensure that these variables always have the value they should have.

[Issue 164](https://github.com/fvwmorg/fvwm3/issues/164)

### DIFF
--- a/fvwm/icccm2.c
+++ b/fvwm/icccm2.c
@@ -51,7 +51,7 @@ SetupICCCM2(Bool replace_wm)
 	XSetWindowAttributes attr;
 	XEvent xev;
 	XClientMessageEvent ev;
-	char wm_sx[20];
+	char wm_sx[25];
 
 	sprintf(wm_sx, "WM_S%lu", Scr.screen);
 	_XA_WM_SX =     XInternAtom(dpy, wm_sx, False);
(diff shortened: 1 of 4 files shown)


ThomasAdam:
Hi @klebertarcisio 

Thanks for this.  How did you derive which variables to change here?  Have you encountered problems where the code had larger values and hence overran?  Or something else?

I'm happy to apply this change -- I'm just wondering how many other instances there might be...

klebertarcisio:
Hi @ThomasAdam,

I used a static analysis tool focused on finding points for improvement. In this specific case, the tool detected a weakness classified as CWE 120. (https://cwe.mitre.org/data/definitions/120.html) (Buffer Copy without Checking Size of Input)
_**Have you encountered problems where the code had larger values and hence overran?**_
We did not find a case that used more memory than was allocated. In fact, this pull request is a matter of prevention.

_**How did you derive which variables to change here?**_
I believe the tool used types to suggest what the real size should be. I believe the tool checks whether "sprintf" and similar operations are working with the correct (or maximum) memory allocation. To verify this, I believe that the tool needs to know the types.

_**I'm happy to apply this change**_
Thanks!

 _**I'm just wondering how many other instances there might be...**_
Since we are dealing with a system that can be used on multiple platforms, it is not so simple to detect all instances. However, I would be happy to send more pull requests in other moments.

PR #307, opened 2020-12-01
FreeBSD: include fvwm_x11.h for modifiers

ThomasAdam:
Add missing header for modifiers.

### DIFF
--- a/libs/modifiers.h
+++ b/libs/modifiers.h
@@ -5,6 +5,7 @@
 
 /* ---------------------------- included header files ---------------------- */
 #include "charmap.h"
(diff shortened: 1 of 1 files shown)


ThomasAdam:
@mikeandmore  -- FYI.  I had to make this change to compile on FreeBSD.

mikeandmore:
Interesting. I think it's because `config.h` still includes a few X11 dependencies. Should we clean that up too?

Either way, this is a good fix.

ThomasAdam:
Hi @mikeandmore 

Sure! If you feel there's a better way, go for it!

PR #324, opened 2020-12-06
freetype: don't clober CFLAGS/LDFLAGS

ThomasAdam:
When detecting Freetype, don't include its own compilation information
directly in CFLAGS/LDFLAGS.  Instead, separate this out.

This fixes the case where a user could override CFLAGS as in:

    make CFLAGS="-O0 -ggdb" -j $(nproc)

yet freetype clobbers this, resulting in freetype not being detected.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -582,6 +582,8 @@ AC_ARG_ENABLE(xft,
   ]
 )
 
+freetype_CFLAGS=""
+freetype_LIBS=""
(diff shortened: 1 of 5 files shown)


ThomasAdam:
@mikeandmore 

FYI -- this is how I solved this.  Seems to work for me.

mikeandmore:
I don't know autoconf very well...but I think this is the best way to fix it. Freetype is just another library so we should handle its cflags in the same way.

PR #343, opened 2020-12-10
configure.ac: update via autoupdate

ThomasAdam:
Fvwm's configure.ac is old, and as such there's newer macros which
should be used, but also older macros which should be deprecated.

'autoupdate' is am autotools program which can help with this.


ThomasAdam:
Not going to bother with this.

Issue #198, opened 2020-09-02
FvwmPager: moving windows inside FvwmPager is inconsistent


ThomasAdam:
When `FvwmPager` became `per-monitor` aware, this worked for scrolling around pages.  But this no longer works very well when moving windows within `FvwmPager` to new desks/pages.

@somiaj has the following logic to fix this:

```
# Definitions
#   nx = Number of Pages in the Horizontal Direction
#   ny = Number of Pages in the Vertical Direction
#
#   px = x-coordinate of window in pager.
#   py = y-coordinate of window in pager.
#
#   PagerWidth = Width of Pager Window
#   PagerHeight = Height of Pager Window
#   (This is for a single desktop)
#
#   DisplayWidth = Width of display (single page)
#   DisplayHeight = Height of display (single page)
#
#   mx = monintor's x-coordinate
#   my = monintor's y-coordinate
#
#   Vx = x-coordinate of current viewport
#   Vy = y-coordinate of current viewport

# Get size of a single page
PageWidth = PagerWidth / nx;
PageHeight = PagerHeight / ny;

# Get current page
Pagex = px / PageWidth;
Pagey = py / PageHeight;

# Get coordinates relative to current page and
# scale coordinates to full display.
cx = (px % PageWidth) * DisplayWidth / PageWidth;
cy = (py % PageHeight) * DisplayHeight / PageHeight;

# Compute display coordinates
x = Pagex * DisplayWidth + cx + mx
y = Pagey * DisplayHeight + cy + my

# Adjust Coordinates to Current Display
t->x = x - Vx
t->y = y - Vy
```

This is probably all via `modules/FvwmPager/x_pager.c:MoveWindow()`

phintsan:
(I don't know if this is the same issue than this one, but for me this is "inconsistent" so I'm adding this report here.)

Here is another sequence in FvwmPager which causes erratic behaviour:

1. Start fvwm3 with the default configuration.
2. Open FvwmConsole.
3. Issue "Module FvwmPager 0 1" command.
4. Drag FvwmConsole window to Desk 1 in the pager. Do this so that the window overlaps the desk boundary and release mouse button on top of Desk 1.

Expected behaviour: FvwmConsole is on Desk 1, partially outsize the virtual screen.
Observed behaviour: FvwmConsole is on Desk 0, partially visible on FvwmPager but invisible on the screen.

I have seen similar mess-ups after other moves between desks, but this is probably the easiest way to reproduce the problem.

[...]

PR #344, opened 2020-12-10
FvwmPager: don't update desk directly

ThomasAdam:
When a window is being moved within the pager, the desk the window is on
was being updated, and then later on, FVWM_UPDATE_SCREEN would further
confuse things.  FvwmPager sends Move/MoveToDesk commands back to Fvwm
to get the window's properties updated.

Let the FVWM_UPDATE_SCREEN logic update the window, since the Move
command FvwmPager sends back will also update the desk the window is on,
so no need to explicitly move the window.

Helps #198


ThomasAdam:
Fixed thanks to other on-going pager work, due in 1.0.4

PR #376, opened 2020-12-22
fix FvwmConsoleC.pl script; fix fvwm-convert-2.6 manpage name

slazav:
1. FvwmConsoleC.pl: remove deprecated defined(%hash) and defined(@array) 

defined(%hash) and defined(@array) are depricated in Perl 5.6.1. "if (%hash)" or "if (@array)" should be used to test if hash or array is not empty.

2. Fix error in manpage name: fvwm-config-2.6 -> fvwm-convert-2.6


ThomasAdam:
Hi @slazav 

Thanks for this. My suggestion is to remove FvwmConsoleC.pl completely.

It's an area of FvwmConsole that isn't used.

Please generate two PRs, one for this change,  and one for the adoc file rename.

ThomasAdam:
Closing due to other PRs.

PR #377, opened 2020-12-22
fix error in manpage name: fvwm-config-2.6 -> fvwm-convert-2.6

slazav:


ThomasAdam:
Hi @slazav 

I agree with this change, however the commit message needs work.  I would suggest something like:

```
doc: fix incorrect naming of fvwm-convert-2.6

When the manpages were converted from Docbook to Asciidoc, the filename for what should have been fvwm-convert-2.6 was misnamed as fvwm-config-2.6.
```

PR #378, opened 2020-12-22
FvwmConsoleC.pl: remove deprecated defined(%hash) and defined(@array)

slazav:


ThomasAdam:
Hi @slazav 

I shan't be merging this PR.  I would much rather see `FvwmConsoleC.pl` removed entirely, rather than these minor perl fixes, since as I said before, `FvwmConsoleC,pl` has not been used for ages.

If you want to do something else that's perl-related, turn your attention to `perllib` and `FvwmPerl`.

slazav:
Thanks. I'll try to prepare a PR for removing FvwmConsoleC.pl. As I understand, it will require some non-trivial changes in FvwmConsole and in documentation.

slazav:
Could you please explain what is written in FvwmConsole.c:123 ?
For me it looks strange. If I understand correctly, you are building arguments for xterm using module arguments in a few loops. In the first loop you process -e and -terminal options, If -e is met you ignore this option and its argument and go to the next loop. Then you add -e FvwmConsoleC and then all options which goes after -e. It means that -terminal before and after -e has different meaning: one changes terminal name, another is passed to the terminal.
Documentation says that -e "is still provided" but "discouraged". In fact argument to -e is ignored, but the option changes order of parameters given to the terminal and meaning of -terminal parameter.

PR #379, opened 2020-12-22
Remove FvwmConsoleC.pl front-end

slazav:
Option -e in FvwmConsole module is still available.


ThomasAdam:
Hi @slazav,

Thanks -- this looks good.  Again though, the commit message needs work.  I would suggest:

```
fvwmconsole: remove older FvwmConsoleC.pl wrapper

When FvwmConsole was first written, there was a companion wrapper written in perl which allowed for a different communication backend to be used with fvwm.

Over time, this was never really used and FvwmConsole itself is being shrunk down to become a compatibility shell for when FvwmPrompt isn't being used.  To reduce the confusion, it's time for FvwmConsoleC.pl to be removed.
```

PR #382, opened 2020-12-22
Fix system.fvwm2rc not found

phileimer:
Since commit 841524a34b5741ae6024e11cf64fc8dcb39057cb, system.fvwm2rc is not found at startup.


ThomasAdam:
Hi @phileimer 

Yeah, OK.  I suppose there's some dusty machines out there which still rely on this...

phileimer:
Hi @ThomasAdam ,
Mine is dusty! ;-)

Issue #333, opened 2020-12-08
Deprecate fvwm's individual screen functionality


ThomasAdam:
Before the days of actual `Xinerama` or `XRandR` integration, it was possible to run `fvwm` on separate displays.   So for example, if there were two physical monitors attached to a computer, then `fvwm` could be started as:

```
fvwm -s 0
fvwm -s 1
```

These days though, RandR support is sufficient such that fvwm shouldn't need to start up on separate displays.  Indeed, a downside to this is that it is not possible for windows to move between displays.

Given the maturity of RandR, this mode in `fvwm3` should be removed.

slazav:
Now if no -s option is given, fork() is done for each screen of a display. Do you want to remove this as well and use only DefaultScreen(dpy)?

ThomasAdam:
Hi @slazav 

That's part of it.  But this is also a lot of work.  It's not as simple as just removing the option for `-s`.  Internally, `fvwm3` is tracking the state of various things which could be operating on separate displays.  For example:

* `LastScreenFocus` -- if you check `fvwm/focus.c` you'll find various uses of that there.  This is involved with *separate* screens.
*  Grep for the string: `pointer is on a different screen` -- this is indicating the pointer is on a different display, and hence sometimes the coordinates are ignored, sometimes the x/y values are set to `0`/`0`.

What we need to do here is carefully evaluate the state-tracking `fvwm3` is doing in these instances and check what the correct behaviour should be.  For that, it will be a little more involved.

@slazav -- I'm more than happy to mentor you through this if you want to give it a go?

slazav:
The first step is clear, removing the -s option and using default screen. This is the first commit: https://github.com/slazav/fvwm3/commit/0990001c9b08389c98d25c81bf000553cb90657f

Interaction with other screens is not clear for me yet. I will try to read the code today-tomorrow.

ThomasAdam:
Hi @slazav,

How are you getting on with this?  Is there anything I can do to help?

slazav:
Hi,

In December I prepared a few commits in this branch:
https://github.com/slazav/fvwm3/tree/remove_screens_%23333
Should I start making a pull request from them?
I can return to this work in a few days, but I'm not sure that I go too deep in it, I'm quite busy at my work now. 

Could you please look at #381, it still does not work for me (should it be reopened?)

ThomasAdam:
Hi @slazav,

[...]

PR #385, opened 2020-12-22
fvwm3: fix segfault caused by incorrect screen number in -s option

slazav:
Way to reproduce the error: call "fvwm3 -s <N>" with some non-existing screen number N. In this case XOpenDisplay() returns dpi==0, program should exit before using it in ScreenCount(dpi)


ThomasAdam:
Hi @slazav,

Per #333 I would rather see work done to remove this, not support it -- even in terms of this bug you've found.

Please feel free to open any PRs which go towards that effort, but I shall close this one for now.

PR #386, opened 2020-12-22
Remove explicit path to bson.h (closes #375)

### LINKED ISSUE #375: libbson detection fails in Linux.
Most Linux distributions use **libbson-1.0/bson.h** instead of bson/bson.h.

The code in fvwm3 uses this:

```
#if defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__sun__)
 #include <libbson-1.0/bson.h>
#else
#include <bson/bson.h>
#endif
```
Which means if not FreeBSD, OpenBSD or Solaris, header include will be **bson/bson.h**. So, the build will fail because there is no /usr/include/bson/bson.h in most Linux distributions.

Removing the conditionals allows to build fvwm3 in Linux.

```
diff -Naur fvwm3-1.0.2.orig/fvwm/builtins.h fvwm3-1.0.2/fvwm/builtins.h
--- fvwm3-1.0.2.orig/fvwm/builtins.h	2020-12-11 18:37:08.000000000 -0600
+++ fvwm3-1.0.2/fvwm/builtins.h	2020-12-21 21:18:36.846795167 -0600
@@ -3,11 +3,7 @@
 #ifndef FVWM_BUILTINS_H
 #define FVWM_BUILTINS_H
 
-#if defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__sun__)
 #include <libbson-1.0/bson.h>
-#else
-#include <bson/bson.h>
-#endif
 
 #include "fvwm.h"
 #include "screen.h"
diff -Naur fvwm3-1.0.2.orig/modules/FvwmMFL/FvwmMFL.c fvwm3-1.0.2/modules/FvwmMFL/FvwmMFL.c
--- fvwm3-1.0.2.orig/modules/FvwmMFL/FvwmMFL.c	2020-12-20 08:47:50.000000000 -0600
+++ fvwm3-1.0.2/modules/FvwmMFL/FvwmMFL.c	2020-12-21 21:18:45.505121652 -0600
@@ -30,11 +30,7 @@
 #include <stdbool.h>
 #include <unistd.h>
 
-#if defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__sun__)
 #include <libbson-1.0/bson.h>
-#else
-#include <bson/bson.h>
-#endif
 
 #include <event2/event.h>
 /* FIXME: including event_struct.h won't be binary comaptible with future
```

slazav:
pkg-config is used in configure script to detect libbson library and to set correct compiler flags (-I...). No need to write system-dependent paths explicitly.

### DIFF
--- a/fvwm/builtins.h
+++ b/fvwm/builtins.h
@@ -3,12 +3,7 @@
 #ifndef FVWM_BUILTINS_H
 #define FVWM_BUILTINS_H
 
-#if defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__sun__)
-#include <libbson-1.0/bson.h>
-#else
-#include <bson/bson.h>
-#endif
-
+#include <bson.h>
 #include "fvwm.h"
 #include "screen.h"
 


ThomasAdam:
Hi @slazav 

As discussed on #375, this change looks good,  However, the commit message needs work.  Please note that PRs to this repository has a template, and I do mention the preferred way of referencing bugs.  For reference, I would want to "closes #XXX" to be at the end of the commit message.  Hence:

```
remove explicit path to bson.h

pkg-config is used in configure script to detect libbson library and to set correct compiler flags.  Allow that to determine the headers paths.

Fixes #375 
```

slazav:
I'm not very accurate in commit messages, sorry that you have to rewrite them.
But for me it's useful to see your variants and easy to change the messages.

PR #402, opened 2021-01-02
Avoid hard coding of version info in configure.ac

pete4abw:
Instead, the shell program, fvwm-version-str.sh will use git describe
and git show to obtain tag and release information from the log.

If the annotated release tag is at the HEAD, the version will be
considered a release. If there are commits after the release tag,
the micro version will have the tag offset as well as the commit
of the HEAD appended to it. Going forward, versions will appear as:

Fvwm3 Configuration:

  Version:     1.0.3 (released)

  Executables: /usr/local/bin
  Man pages:   /usr/local/share/man
  Modules:     /usr/local/libexec/fvwm3/1.0.3

for a Release, and

Fvwm3 Configuration:

  Version:     1.0.2-25-b9435cd (not released)

  Executables: /usr/local/bin
  Man pages:   /usr/local/share/man
  Modules:     /usr/local/libexec/fvwm3/1.0.2-25-b9435cd

when HEAD commit <> release tag commit. This example indicated there
are 25 commits after the last release.

The program fvwm-version-str.sh can be run independently to test
the impact of different scenarios. It is fuly documented and may be
altered to taste.

The file, configure.ac currently searches for annotated tags, not
lightweight tags. Removing the option -r from the m4_esyscmd_s command
will allow it to search for the latest tag.

m4_define([v_maj],		[m4_esyscmd_s([./utils/fvwm-version-str.sh major -r])])

On branch ph/gitdesc
	modified:   Makefile.am
	modified:   configure.ac
	modified:   utils/fvwm-version-str.sh

* **What does this PR do?**

    the PR can be merged.

* **Issue number(s)**

message(s) contains:

```
Fixes #XXX
```

relevant issue number.

```
```


utils/fvwm-version-str.sh
```diff
@@ -10,12 +10,180 @@
...
+	fi
+
+	# assign minor version
+	minor=$(echo $version | cut -d. -f2)
+
+	# assign major version
+	major=$(echo $version | cut -d. -f1)
+}
+
+[ ! $(which git) ] && die "Something very wrong: git not found."
```
ThomasAdam, line 130:
```
type -p git >/dev/null || die "....."
```

utils/fvwm-version-str.sh
```diff
@@ -10,12 +10,180 @@
...
+# with leading v and commit leadint g removed
+# tag commit dates obtained with git show
+
+init() {
+	local r_describe
+	# git describe raw format
+	describe_tag=$(git describe $tagopt --long --abbrev=$commit_length)
+
+	# git describe prefixes commit with the letter `g'
+	describe_tag=${describe_tag/g/}
```
ThomasAdam, line 98: Ah.  Please don't do that.  The "g-" prefix is useful and I use it all the time to run `git checkout` on it.  Removing this "g" breaks that.  It's actually there for a reason.  From `git-describe(1)`:

```
       The number of additional commits is the number of commits which would be displayed by "git log v1.0.4..parent". The hash suffix is "-g" + unambiguous abbreviation for the tip commit of parent
       (which was 2414721b194453f058079d897d13c4e377f92dc6). The "g" prefix stands for "git" and is used to allow describing the version of a software depending on the SCM the software is managed with.
       This is useful in an environment where people may use different SCMs.
```

ThomasAdam (changes requested):
Hi @pete4abw 

Thanks for sending this over.  The PR itself is still marked as in-progress, so am not sure if you're expecting me to fully review it yet, but I've left some comments for you nonetheless.

It currently fails to build -- you might want to check the to see why.

pete4abw:
**Thank you** for the review, @ThomasAdam . This is what I needed since I am new in the project.

I added quoting for VERSIONINFO in configure.ac Compiles now
Removed licensing for fvwm_version_str.sh. Add whatever you need.
Switched from using `which` to `type -p`
Restored leading `g` to commit value

I was confused by the AC_DEFINE_UNQUOTED. It messed up the created `config.h` file where
**VERSIONINFO**=not released instead of "not released"
with no quotes.

Have another looksee!

PR #411, opened 2021-01-20
Remove screens #333

### LINKED ISSUE #333: Deprecate fvwm's individual screen functionality

slazav:
Removing multi-display support from fvwm
Fixes #333


ThomasAdam:
Thanks, @slazav -- I'll pick this up and finish it off.

gerddie:
Do I understand correctly that this MR intends to remove the possibility to run fvwm3 with a setup using separate X-screens (i.e. where I can switch work spaces separately on one or the other screen)?  The reason why I'm currently trying out fvwm3 is exactly this, because most windows managers don't seem to support this anymore, and it would be kind of sad if fvwm3 would also remove this option.

ThomasAdam:
> Do I understand correctly that this MR intends to remove the possibility to run fvwm3 with a setup using separate X-screens (i.e. where I can switch work spaces separately on one or the other screen)?  The reason why I'm currently trying out fvwm3 is exactly this, because most windows managers don't seem to support this anymore, and it would be kind of sad if fvwm3 would also remove this option. 

Hi @gerddie 

No, this is not what this refers to. What you want is the setting `DesktopConfiguration per-monitor` which isn't being removed.

What this PR refers to is having separate physical screens to run multiple FVWM instances on, which are separate to one another.

Issue #418, opened 2021-01-31
Invalid read of size 1 (reading freed memory)


Quipyowert2:
Thanks for reporting your bug here!  The following template will help with
giving as much information as possible so that it's easier to diagnose and
fix.

## Upfront Information

Please provide the following information by running the command and providing
the output.

* Fvwm3 version (run: `fvwm3 --version`)
```
fvwm3 1.0.3 (1.0.2-48-g356e327f)
with support for:  ReadLine, XPM, PNG, Shape, XShm, SM, Bidi text, XRandR, XRender, XCursor, XFT, NLS
```
* Linux distribution or BSD name/version
openSUSE 15.2 on Windows Subsystem for Linux
* Platform (run: `uname -sp`)
Linux x86_64
Valgrind version:
valgrind-3.15.0
## Expected Behaviour

What were you trying to do?  Please explain the problem.

Ideally, no errors reported by Valgrind.

## Actual Behaviour

Valgrind reported an "Invalid read of size 1" when `__handle_bpress_on_managed` accessed the `(fw)->flags.is_scheduled_for_raise` flag of a window that was freed during a call to `__handle_bpress_action` (by CMD_close).

What should have happened, but didn't?
Fvwm should avoid accessing freed memory.

Valgrind output:
```
==23550== 
==23550== Invalid read of size 1
==23550==    at 0x42E26B: __handle_bpress_on_managed (fvwm/events.c:1727)
==23550==    by 0x42E26B: HandleButtonPress (fvwm/events.c:1843)
==23550==    by 0x42C1D9: dispatch_event (fvwm/events.c:4146)
==23550==    by 0x42CFF7: HandleEvents (fvwm/events.c:4194)
==23550==    by 0x40B74A: main (fvwm/fvwm3.c:2560)
==23550==  Address 0x98be0f0 is 336 bytes inside a block of size 1,312 free'd
==23550==    at 0x4C2F50B: free (in /usr/lib64/valgrind/vgpreload_memcheck-amd64-linux.so)
==23550==    by 0x44161F: destroy_window (fvwm/add_window.c:3518)
==23550==    by 0x43BC85: CMD_Close (fvwm/builtins.c:2485)
==23550==    by 0x4644E9: __execute_function.part.2 (fvwm/functions.c:626)
==23550==    by 0x42E84B: __handle_bpress_action (fvwm/events.c:1623)
==23550==    by 0x42E84B: __handle_bpress_on_managed (fvwm/events.c:1721)
[...]

PR #419, opened 2021-01-31
Fix invalid read of size 1 error from Valgrind.

Quipyowert2:
After __handle_bpress_action is called, the window fw might have been
freed by CMD_close. Check that the window still exists before checking
if it is scheduled to be raised, so that we don't attempt to read freed memory.

* **What does this PR do?**
Fixes an attempt to read freed memory. (#418)

    the PR can be merged.

* **Issue number(s)**

message(s) contains:

```
Fixes #XXX
```

relevant issue number.

```
```

### DIFF
--- a/fvwm/events.c
+++ b/fvwm/events.c
@@ -1724,7 +1724,7 @@ static void __handle_bpress_on_managed(const exec_context_t *exc)
 		}
 	}
 	/* raise the window */
-	if (IS_SCHEDULED_FOR_RAISE(fw))
+	if (check_if_fvwm_window_exists(fw) && IS_SCHEDULED_FOR_RAISE(fw))
 	{
 		/* Now that we know the action did not restack the window we
 		 * can raise it.


fvwm/events.c
```diff
@@ -1724,7 +1724,7 @@ static void __handle_bpress_on_managed(const exec_context_t *exc)
 		}
 	}
 	/* raise the window */
-	if (IS_SCHEDULED_FOR_RAISE(fw))
+	if (check_if_fvwm_window_exists(fw) && IS_SCHEDULED_FOR_RAISE(fw))
```
ThomasAdam, line 1727: Since `check_if_fvwm_window_exists()` just walks to the `FvwmWindow` structure looking for itself, we don't need to do this, and can instead just check what `fw` is, as in:

ThomasAdam (changes requested):
Hi @Quipyowert2 

Thanks.  This seems like a false positive to me.  Before calling `__handle_bpress_on_managed()` we've already checked to see if the window is valid (i.e., not NULL).  For it to go between then, and raising the window, `IS_SCHEDULED_FOR_DESTROY` would have to happen.

Since it's very unlikely, we can still leave a check in, I suppose.  I've left a suggestion at the appropriate line.

fvwm/events.c
Quipyowert2: I compiled and ran fvwm3 from master branch (356e327f) with the `vgdb` gdb server (`valgrind --vgdb-error=0 fvwm/fvwm3` and then in another terminal `gdb -ex 'target remote | vgdb'` and then `c`ontinue past the uninitialized value errors) and after Valgrind prints the Invalid read of size 1 error (which happens when closing a window) gdb says `fw` is 0x947bd60 and `monitor get_vbits 0x947bd60` (which asks Valgrind about the validity of an address) says that fw is unaddressable.

I don't think a NULL check would fix the problem, because `fw` has been freed but hasn't been set to NULL, so it's a dangling pointer. Another way to solve this would be to use a function that takes a pointer to pointer and free the pointed-to pointer and then set it to NULL. Something like [this mem_dealloc_ function](https://codereview.stackexchange.com/questions/26656/are-these-memory-allocation-wrapper-functions-kosher-with-all-c-compilers).

`destroy_window()` seems to remove `fw` from the window list at [addwindow.c:3367](https://github.com/fvwmorg/fvwm3/blob/356e327f16b40effe859b517d0d486e998146283/fvwm/add_window.c#L3367). If `Scr.FvwmRoot` in check_if_fvwm_window_exists is part of the same linked list then `fw` won't be in the list so the condition at line 1727 of events.c will short-circuit and not check `IS_SCHEDULED_FOR_RAISE(fw)` thereby avoiding the use-after-free.

Valgrind doesn't report the invalid read error if I compile fvwm3 from my branch nm/fix-invalid-read and repro with the steps at issue #418.

I had to recompile events.c  with -O0 to get gdb to tell me the actual value of `fw` instead of saying `<optimized out>`

GDB output after valgrind hits the invalid read error with master branch (356e327f ) of fvwm3:
```
Program received signal SIGTRAP, Trace/breakpoint trap.
0x000000000042b8dd in __handle_bpress_on_managed (exc=0x948ed70) at events.c:1727  
1727            if (IS_SCHEDULED_FOR_RAISE(fw))
(gdb) print fw
$1 = (FvwmWindow * const) 0x947bd60
(gdb) monitor get_vbits 0x947bd60
__ 
Address 0x947BD60 len 1 has 1 bytes unaddressable 
(gdb)
```

ThomasAdam: Hi @Quipyowert2 

Thanks.   Please can you include this in your ammended commit message in your PR, and I'll approve it.  Well spotted!

ThomasAdam:
Thanks again, @Quipyowert2.

PR #423, opened 2021-02-03
Remove package (libx11-dev) already present in core dependencies

djwf:
I figured this did not need an issue of its own, but let me know if you would like me to file one.

### DIFF
--- a/dev-docs/INSTALL.md
+++ b/dev-docs/INSTALL.md
@@ -37,7 +37,6 @@ FVWM3 has the following dependencies.
 * libreadline-dev
 * librsvg-dev
 * libsm-dev
-* libx11-dev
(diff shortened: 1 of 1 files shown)


ThomasAdam:
Hi @djwf 

Not sure what system you're using, but the X11 headers are most definitely not present on all systems by default, so I'm wary of removing this.

djwf:
Hi @ThomasAdam: this is removing the libx11-dev from the list of optional dependencies, not from the core dependencies (that package is present in both lists). Sorry, I should have been more clear in the description.

ThomasAdam:
Hi @djwf 

Oh right, I see.  Yep, you're right.  Will merge.

PR #428, opened 2021-02-07
Fix several tiny memory leaks

Quipyowert2:
* **What does this PR do?**
Fixes a series of very tiny leaks.

    the PR can be merged.

* **Issue number(s)**

message(s) contains:

```
Fixes #XXX
```

relevant issue number.

```
```

### DIFF
--- a/fvwm/virtual.c
+++ b/fvwm/virtual.c
@@ -1738,7 +1738,7 @@ Bool get_page_arguments(FvwmWindow *fw, char *action, int *page_x, int *page_y, 
 	int suffix[2];
 	int mw, mh;
 	char *token;
-	char *taction, *action_cpy;
+	char *taction, *action_cpy, *action_cpy_start;
 	int wrapx;
 	int wrapy;
 	int limitdeskx;
@@ -1752,7 +1752,9 @@ Bool get_page_arguments(FvwmWindow *fw, char *action, int *page_x, int *page_y, 
 
 	m_use = (fw && fw->m) ? fw->m : monitor_get_current();
 	action_cpy = strdup(action);
+	action_cpy_start = action_cpy;
 	token = PeekToken(action_cpy, &action_cpy);
+	free(action_cpy_start);
 
 	if (token == NULL)
 		return (False);
@@ -2332,9 +2334,10 @@ void CMD_DesktopSize(F_CMD_ARGS)
 void CMD_GotoDesk(F_CMD_ARGS)
 {
 	struct monitor  *m_use = monitor_get_current(), *m;
-	char		*action_cpy, *token;
+	char		*action_cpy, *action_cpy_start, *token;
 
 	action_cpy = strdup(action);
+	action_cpy_start = action_cpy;
 	token = PeekToken(action_cpy, &action_cpy);
 
 	m = monitor_resolve_name(token);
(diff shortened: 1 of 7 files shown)


libs/FScreen.c
```diff
@@ -409,9 +409,13 @@ scan_screens(Display *dpy)
 	for (i = 0; i < n; i++) {
 
 		char *name = XGetAtomName(dpy, rrm[i].name);
+		Bool must_free_name = True;
```
ThomasAdam, line 412: As an aside here, the "Bool" type is usually reserved for Xlib operations, and either "int" or "bool" (via `stdbool.h`) can be used instead.

libs/FScreen.c
```diff
@@ -409,9 +409,13 @@ scan_screens(Display *dpy)
...
 
 		if (name == NULL)
+		{
 			name = "";
+			must_free_name = False;
+		}
```
ThomasAdam, line 419: Looking at this a bit more, if we can't detect a monitor name coming from the XAtom, then it's an error, and not something we should continue with.  In detecting monitors, we won't ever have stored the empty string to match against, so the rest of the function won't really do a great deal.  I'd rather see:

```
if (name == NULL) {
    fprintf(stderr, "%s: couldn't detect monitor with empty name\n", __func__);
    exit (101);
}
```

libs/FScreen.c
```diff
@@ -447,9 +451,12 @@ scan_screens(Display *dpy)
 		m->virtual_scr.MyDisplayWidth = monitor_get_all_widths();
 		m->virtual_scr.MyDisplayHeight = monitor_get_all_heights();
 
+		if (must_free_name)
+			XFree(name);
```
ThomasAdam, line 455: No need to check to free here, given the above change.

libs/FScreen.c
```diff
@@ -447,9 +451,12 @@ scan_screens(Display *dpy)
...
 	}
 
 	monitor_check_primary();
+	XRRFreeMonitors(rrm);
```
ThomasAdam, line 459: This is correct.

ThomasAdam (changes requested):
Hi @Quipyowert2 

Thanks, all look good.  Not overly wild about the whole "Bool" checking in functions.c, but there's already a precendence for it.

See comments in-line for the FScreen.c change.

Quipyowert2:
I added the error message you suggested and removed the `must_free_name` variable since it wasn't needed. I squashed the commits into one commit but the commit message is kind of long.

ThomasAdam:
Thanks, @Quipyowert2 

I would use a commit message of:

```
Assorted mem leak fixes

In running fvwm3 though valgrind, this commit fixes a few memory leaks it has detected.
```

fvwm/virtual.c
```diff
@@ -2383,7 +2388,7 @@ void CMD_GotoDeskAndPage(F_CMD_ARGS)
 	else
 		PeekToken(action, &action);
 
-	//free(action_cpy);
+	free(action_cpy_start);
```
ThomasAdam, line 2391: Oops.  Thanks for fixing this!

Quipyowert2: `PeekToken` moves `action_cpy` forward a few bytes so I had to introduce another variable to be able to free the original pointer. Freeing `action_cpy` directly caused an Invalid free error from valgrind since it was pointing into the middle of the allocated memory instead of the beginning.

You're welcome.

PR #431, opened 2021-02-14
Fix a 21 byte memory leak.

### LINKED ISSUE #430: 21 byte memory leak in merge_styles (3 tiny leaks of 7 bytes)
Thanks for reporting your bug here!  The following template will help with
giving as much information as possible so that it's easier to diagnose and
fix.

## Upfront Information

the output.

4405437b 
```
fvwm3 1.0.3 (1.0.2-48-g356e327f)
with support for:  ReadLine, XPM, PNG, Shape, XShm, SM, Bidi text, XRandR, XRender, XCursor, XFT, NLS
```
openSUSE 15.2 running on Windows Subsystem for Linux.
* Platform (run: `uname -sp`)
Linux x86_64 
## Expected Behaviour
Ideally no memory leaks reported by Valgrind
What were you trying to do?  Please explain the problem.
Running valgrind on fvwm3 to find any more memory leaks to fix.
## Actual Behaviour
[merge_styles_leak.log](https://github.com/fvwmorg/fvwm3/files/5976833/merge_styles_leak.log)

What should have happened, but didn't?

## Enabling logging

`fvwm3` has a means of logging what it's doing.  Enabling this when
reproducing the issue might help.  To do this, either change the means fvwm3
is started by adding `-v` as in:

```
fvwm3 -v
```

or, once `fvwm3` has loaded, send `SIGUSR2` as in:

```
pkill -USR2 fvwm3
```

The resulting logfile can be found in `$HOME/.fvwm/fvwm3-output.log`

## Steps to Reproduce

How can the problem be reproduced? 
1. Compile fvwm3 revision 4405437b
2. `valgrind --leak-check=full --track-origins=no --undef-value-errors=no --fullpath-after=$PWD/ --num-callers=200 --suppressions=fontconfig.supp fvwm/fvwm3 |& less`
3. When Fvwm3 loads, click the background and click Quit then click Quit again from the confirmation dialog that pops up.
For this, the following is helpful:

* Reduce the problem to the smallest `fvwm` configuration example (where
  possible).  Start with a blank config file (`fvwm3 -f/dev/null`) and go from
  there.

Include your configuration with this issue.
My fvwm config is the default fvwm3 config with this line added from a bugreport somewhere (I think from PR #429):
`*FvwmIconMan*action      Mouse   1 N sendcommand "Iconify off", sendcommand "FlipFocus", sendcommand "Raise"`
## Does Fvwm3 crash?
Yes, it crashed several times when I was adding in the free_style calls but seems to be fine now.

If fvwm3 crashes then check your system for a corefile.  This is platform
dependant, however, check if:

  - corefiles are enabled (`ulimit -c`)
  - A corefile might be in `$HOME` or `/tmp/`.
  - If you're using Linux (with `systemd`), check `coredumpctl list`.
    `coredumpctl` may need installing separately.

If you find a corefile, install `gdb` and run:

```
gdb /path/to/fvwm3 /path/to/corefile
```

If you're using `coredumpctl` then use:

```
coredumpctl debug
```

Then from within the `(gdb) ` prompt, issue:

```
bt full
```

... and include the output here.

## Extra Information

X server: VcXsrv 1.20.8.1
* Feel free to take a screen capture or video and upload to this issue if you
  feel it would help.

* Attach `$HOME/.fvwm/fvwm3-output.log` from the step above.
[fvwm3-output.log](https://github.com/fvwmorg/fvwm3/files/5976836/fvwm3-output.log)

Fontconfig error suppression file (`fontconfig.supp` in the above Valgrind command):
```
{
   Fontconfig suppression
   Memcheck:Leak
   match-leak-kinds: definite
   fun:malloc
   ...
   obj:/usr/lib64/libfontconfig.so.1.10.1
   ...
}
{
   Fontconfig suppression 2
   Memcheck:Leak
   match-leak-kinds: definite
   fun:realloc
   ...
   obj:/usr/lib64/libfontconfig.so.1.10.1
   ...
}
```

Quipyowert2:
merge_styles allocated a small amount of memory using strdup(). This
memory wasn't being freed by the callers. I had to change
do_free_src_and_alloc_copy argument of merge_styles to True in
lookup_style() to get free_style() to stop crashing.

Fixes #430.

* **What does this PR do?**
Fixes a memory leak by always calling merge_styles with do_free_src_and_alloc_copy set to True and then calling free_style on the style later.

    the PR can be merged.

* **Issue number(s)**

message(s) contains:

```
Fixes #XXX
```

relevant issue number.

```
```

### DIFF
--- a/fvwm/add_window.c
+++ b/fvwm/add_window.c
@@ -1828,6 +1828,7 @@ int setup_visible_names(FvwmWindow *fw, int what_changed)
 			free(ext_name);
 		}
 	}
+	free_style(&style);
 
 	return (changed_styles) ? changed_styles : affected_titles;
 }
@@ -2358,6 +2359,7 @@ FvwmWindow *AddWindow(
 		}
 		free(fw);
 		MyXUngrabServer(dpy);
+		free_style(&style);
 		return NULL;
 	}
 
@@ -2389,6 +2391,7 @@ FvwmWindow *AddWindow(
 		}
 		free(fw);
 		MyXUngrabServer(dpy);
+		free_style(&style);
 		return AW_UNMANAGED;
 	}
 
@@ -2807,6 +2810,7 @@ FvwmWindow *AddWindow(
 		destroy_window(fw);
 		fw = NULL;
 	}
+	free_style(&style);
 
 	return fw;
 }
--- a/fvwm/placement.c
+++ b/fvwm/placement.c
@@ -2426,6 +2426,7 @@ void CMD_PlaceAgain(F_CMD_ARGS)
 		AnimatedMoveFvwmWindow(
 			fw, FW_W_FRAME(fw), -1, -1, attr_g.x, attr_g.y, False,
 			-1, ppctMovement);
+		free_style(&style);
 	}
 	if (fw->Desk != old_desk)
 	{
--- a/fvwm/style.c
+++ b/fvwm/style.c
@@ -785,7 +785,7 @@ static void merge_styles(
 	return;
 }
 
-static void free_style(window_style *style)
+void free_style(window_style *style)
 {
 	/* Free contents of style */
 	SAFEFREE(SGET_NAME(*style));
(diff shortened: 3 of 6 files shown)


Quipyowert2:
Pushed to add `free_style` calls in `CMD_PlaceAgain` and `desk_get_fw_count` near where they call lookup_style.

PR #444, opened 2021-02-23
fix with some pixels off for 3d buttons

mikeandmore:
Some pixels are off with 3d buttons. I believe when drawing buttons a = 1 so we are missing out a pixel by adding `l`.


ThomasAdam:
Thanks, @mikeandmore  -- this looks good to me.  I'll merge this and let's see if anyone else screams.

Issue #446, opened 2021-02-23
Font-rendering causes fvwm to lock-up


ThomasAdam:
When starting fvwm3 (or restarting it), often times fvwm3 will lock-up and become unresponsive until killed.  gdb output below:

```
(gdb) bt                                                                                        
#0  0x00007f9c53a87950 in XftCharIndex () from /usr/lib/libXft.so.2                                                                                                                              
#1  0x00007f9c53a8400f in XftTextExtents32 () from /usr/lib/libXft.so.2                                                                                                                          
#2  0x000055e894ef4595 in MatchFont (dpy=0x55e895ddd300, font=0x55e895ecd280, xoff=0x7ffe52a061e0, yoff=0x7ffe52a061e4,                                                                          
    p=0x55e895f0feb0 "[0x1a00005] fix with some pixels off for 3d buttons by mikeandmore \267 Pull Request #444 \267 fvwmorg/fvwm3 - Google Chrome (1)", len=122, render_char=0x0, user_data=0x0)
 at Fft.c:337
#3  0x000055e894ef4ef6 in FftTextWidth (dpy=0x55e895ddd300, flf=0x55e895ebfa00, str=0x55e895f0feb0 "[0x1a00005] fix with some pixels off for 3d buttons by mikeandmore \267 Pull Request #444 \26
7 fvwmorg/fvwm3 - Google Chrome (1)", len=122)
    at Fft.c:502                                                                                
#4  0x000055e894ed673f in FlocaleTextWidth (flf=0x55e895ebfa00, str=0x55e895f0feb0 "[0x1a00005] fix with some pixels off for 3d buttons by mikeandmore \267 Pull Request #444 \267 fvwmorg/fvwm3 
- Google Chrome (1)", sl=122)
    at Flocale.c:2214                                                                           
#5  0x000055e894e49b45 in border_get_titlebar_descr (cd=0x7ffe52a06600, fw=0x55e895f113d0, pressed_parts=PART_NONE, pressed_button=-1, force_draw_parts=PART_TITLEBAR, clear_parts=CLEAR_ALL, old
_g=0x55e895f11698, new_g=0x55e895f11698, 
    do_hilight=0, ret_td=0x7ffe52a06310) at borders.c:4265                                      
[...]

PR #447, opened 2021-02-23
fix uninitialized value and bad free

mikeandmore:
fixed some heap corruption. 

Some of them can be triggered by start/restart fvwm.

### DIFF
--- a/fvwm/add_window.c
+++ b/fvwm/add_window.c
@@ -407,7 +407,7 @@ static void setup_window_structure(
 	  also need tweaking, depending on what you want to preserve.
 	  For now, just zap any old information, except the desk.
 	*/
-	memset(*pfw, '\0', sizeof(FvwmWindow));
+	memset(*pfw, 0, sizeof(FvwmWindow));
 	FW_W(*pfw) = w;
 	if (savewin != NULL)
 	{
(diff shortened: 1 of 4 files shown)


fvwm/add_window.c
```diff
@@ -1819,7 +1819,7 @@ int setup_visible_names(FvwmWindow *fw, int what_changed)
 			&bits, fw, &style, True);
 		if ((changed_names & bits) || (force_update & 2))
 		{
-			free(fw->visible_icon_name);
+			if (fw->visible_icon_name != NoName) free(fw->visible_icon_name);
```
ThomasAdam, line 1822: Style-nit:

```
if (fw->visible_icon_name != NoName)
    free(fw->visible_icon_name);
```

libs/FEvent.c
```diff
@@ -897,6 +897,7 @@ Status FGetWMNormalHints(
 {
 	Status ret;
 
+	memset(hints_return, 0, sizeof(XSizeHints));
```
ThomasAdam, line 900: I doubt this is strictly necessary?

ThomasAdam (changes requested):
Hey @mikeandmore 

Thanks.  One style-nit I'd like you to fix, and a query regarding a `memset()`...

Can you amend the commit message, add:

```
Fixes #446 
```

libs/FEvent.c
mikeandmore: it is. we kinda ignore the hint masks.

fvwm/add_window.c
mikeandmore: fixed.

mikeandmore:
> ```
> Fixes #446 
> ```

Does it actually fix that? I couldn't reproduce #446. Does this patch work for you?

mikeandmore:
This commit fixes #446

PR #448, opened 2021-02-24
add a CursorSet command

mikeandmore:
`CursorSet` can set an absolute mouse cursor position. This is helpful for save and restoring mouse cursor positions.

On IRC, we talked about support some auto mouse move features from CWM. With `CursorSet`, you no longer need an external script/process and `CursorMove` to save and restore the cursor.

### DIFF
--- a/fvwm/builtins.c
+++ b/fvwm/builtins.c
@@ -2286,14 +2286,21 @@ void CMD_CursorMove(F_CMD_ARGS)
 	int x_pages, y_pages;
 	struct monitor	*m = NULL;
 
-	if (GetTwoArguments(action, &val1, &val2, &val1_unit, &val2_unit) != 2)
-	{
+	if (GetTwoArguments(action, &val1, &val2, &val1_unit, &val2_unit) != 2) {
 		fvwm_debug(__func__, "CursorMove needs 2 arguments");
 		return;
 	}
-	if (FQueryPointer(dpy, Scr.Root, &JunkRoot, &JunkChild,
-			  &x, &y, &JunkX, &JunkY, &JunkMask) == False)
-	{
+	if (val1 == 0 && val2 == 0) {
+		/* Move to absolute point */
+		action = SkipNTokens(action, 2);
+		if (GetTwoArguments(action, &val1, &val2, &val1_unit, &val2_unit) != 2) {
+			return;
+		}
+		/* x and y are 0 and we skip the QueryPointer so that we can
+		 * move to the absolute point. */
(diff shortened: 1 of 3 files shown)


fvwm/builtins.c
```diff
@@ -2385,6 +2385,55 @@ void CMD_CursorMove(F_CMD_ARGS)
...
+		fvwm_debug(__func__, "CursorSet needs 2 arguments");
+		return;
+	}
+	int x = val1 * val1_unit / 100;
+	int y = val2 * val2_unit / 100;
+	unsigned int w = 0, h = 0;
+	struct monitor *m = NULL;
+	TAILQ_FOREACH(m, &monitor_q, entry) {
+		w = m->virtual_scr.MyDisplayWidth;
+		h = m->virtual_scr.MyDisplayHeight;
```
ThomasAdam, line 2403: This needs correcting, as it's been a long-standing semantic bug.  `m->virtual_scr.MyDisplayWidth` and `m->virtual_scr.MyDisplayHeight` are not per-monitor values, but an aggregate of all displays connected.  Hence, you either want to set `w`/`h` to `monitor_get_all_widths()`/`monitor_get_all_heights()` respectively outside of the loop, or use `m->si->w`/`m->si->h` for per-monitor values.

I'll clean this up separately.

ThomasAdam (changes requested):
Hi @mikeandmore 

I suppose the idea makes sense, but I really don't want a separate command for this.  If it were going to be included in `fvwm3` then I'd want it part of `CursorMove` somehow.  The long-term goal is to reduce commands, not increase them.

You'll also need to add to the man page.

mikeandmore:
let me work on adding this into `CursorMove`. In that case i'll reuse the code left there before.

mikeandmore:
This commit will reuse the `CursorMove` command.

mikeandmore:
I separate the commit this time.

ThomasAdam:
Thanks.  Now we can clean things up a bit further:

* Since commit 6cc71636cfbf839aeca2107a84026da3cea66585 was introduced by 728e21a67b8ed6a4f16556f8766ed95eff67aece, please squash 6cc71636cfbf839aeca2107a84026da3cea66585 into 728e21a67b8ed6a4f16556f8766ed95eff67aece
* We can probably reword the commits a  better.  For example, I would word cf85ff991100c9a1dcfdd2cc2cb7bde7800a0294 as:
```
GetTwoPercentArguments: remove unneeded PeekToken

Remove an unneeded call to PeekToken as this is never used.
```

PR #450, opened 2021-02-24
perms: remove erroneous 644 -> 755

ThomasAdam:
Some source files were committed with a change of 644 to 755, and no
execute bit is required on these files.


ThomasAdam:
@Quipyowert2 

FYI, I noticed that some of the source files had changed to having execute bits on them.    Turns out this was happening in your PRs -- I suspect you've something interesting happening locally before you `git push` to Github which causes this.

Please check.

Quipyowert2:
I think the permissions changed because I was using gvim.exe (GUI vim for Windows) to edit the files instead of WSL vim. `gvim.exe` doesn't know about WSL permissions so it saved files with the wrong permissions.

PR #453, opened 2021-03-02
change the default of warp to window to 50%,50%

mikeandmore:
If no parameter is given to `WarpToWindow`, then it defaults to `WarpToWindow 0 0` without reporting an error.

I believe this is intentional, because `WarpToWindow` with two variables that fail to expand can cause this undocumented behaviour. However, 0, 0 means at the top-left of the window, which is extremely awkward. This patch changes this (undocumented) default to 50%, 50%, which is the centre of the window.

### DIFF
--- a/fvwm/focus.c
+++ b/fvwm/focus.c
@@ -1252,30 +1252,35 @@ void CMD_WarpToWindow(F_CMD_ARGS)
 		}
 		else
 		{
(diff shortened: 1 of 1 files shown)


ThomasAdam:
Hi @mikeandmore 

This sort of thing is already configurable, and been this way for ages.  I don't see much point in changing it now.

PR #454, opened 2021-03-02
Revert the commit that supports CursorMove to an absolute position

mikeandmore:
This PR reverts the `CursorMove 0,0,x,y` usage because `WarpToWindow` provides exactly the same thing.

Also update the document in case another user is looking for the same functionality.


ThomasAdam (changes requested):
Hi @mikeandmore 

Looks good.  Commit message needs changing though:

```
doc: mention WarpToWindow when trying to move the cursor on root window

WarpToWindow already allows for the mouse cursor to be moved to an absolute position on the screen. Mention this in context of the CursorMove command so it's more discoverable to users.
```

The point here is to make these commit messages more readable for the future.

mikeandmore:
done.

Issue #107, opened 2020-05-29
Code Cleanup: Codacy issues list


ThomasAdam:
Codacy is a free code analysis tool which will help identify security problems in code.

The current list of issues found against `master` is here:

https://app.codacy.com/gh/fvwmorg/fvwm3/issues/index

It would be good if some/all of these are addressed via PRs.  Some are easier to fix than others, so some of the suggestions might need discussion.

PR #471, opened 2021-03-30
Fix potential buffer overflow in session handling code.

Quipyowert2:
This buffer overflow is fixed by including field width limits in the format strings passed to `sscanf`.

sscanf's %s and %[ format specifiers can overflow the output buffer unless a
field width limit is given. if sscanf encounters a string without
spaces (with %s) or without the characters in the set given (with %[) and this
string is longer than the output buffer, it will overflow the buffer.

* **What does this PR do?**
Fixes potential buffer overflow in session handling code.

Cppcheck 2.4.1 warnings:
```
fvwm\session.c:376:4: warning: sscanf() without field width limits can crash with huge input data. [invalidscanf]
   sscanf(s, "%*s %[^\n]", s1);
   ^
fvwm\session.c:1225:4: warning: sscanf() without field width limits can crash with huge input data. [invalidscanf]
   sscanf(s2, "%[^\n]", s1);
   ^
fvwm\session.c:1236:4: warning: sscanf() without field width limits can crash with huge input data. [invalidscanf]
   sscanf(s2, "%[^\n]", s1);
   ^
fvwm\session.c:1350:4: warning: sscanf() without field width limits can crash with huge input data. [invalidscanf]
   sscanf(s, "%*s %s", s1);
   ^
fvwm\session.c:1469:4: warning: sscanf() without field width limits can crash with huge input data. [invalidscanf]
   sscanf(s2, "%[^\n]", s1);
   ^
fvwm\session.c:1479:4: warning: sscanf() without field width limits can crash with huge input data. [invalidscanf]
   sscanf(s2, "%[^\n]", s1);
   ^
fvwm\session.c:1489:4: warning: sscanf() without field width limits can crash with huge input data. [invalidscanf]
   sscanf(s2, "%[^\n]", s1);
   ^
fvwm\session.c:1499:4: warning: sscanf() without field width limits can crash with huge input data. [invalidscanf]
   sscanf(s2, "%[^\n]", s1);
   ^
fvwm\session.c:1509:4: warning: sscanf() without field width limits can crash with huge input data. [invalidscanf]
   sscanf(s2, "%[^\n]", s1);
   ^
fvwm\session.c:1523:5: warning: sscanf() without field width limits can crash with huge input data. [invalidscanf]
    sscanf (s+pos, "%s%n", s1, &pos1);
    ^
```

Cppcheck warning explanation:
```
CWE: 119
sscanf() without field width limits can crash with huge input data. Add a field width specifier to fix this problem.

Sample program that can crash:

#include <stdio.h>
int main()
{
    char c[5];
    scanf("%s", c);
    return 0;
}

Typing in 5 or more characters may make the program crash. The correct usage here is 'scanf("%4s", c);', as the maximum field width does not include the terminating null byte.
Source: http://linux.die.net/man/3/scanf
Source: http://www.opensource.apple.com/source/xnu/xnu-1456.1.26/libkern/stdio/scanf.c
```

* **Issue number(s)**
#107 
message(s) contains:

```
Fixes #XXX
```

relevant issue number.

```
```

### DIFF
--- a/fvwm/session.c
+++ b/fvwm/session.c
@@ -373,7 +373,7 @@ static Bool VerifyVersionInfo(char *filename)
 		if (!strcmp(s1, "[FVWM_VERSION]"))
 		{
 			char *current_v = get_version_string();
-			sscanf(s, "%*s %[^\n]", s1);
+			sscanf(s, "%*s %4095[^\n]", s1);
 			if (strcmp(s1, current_v) == 0)
 			{
 				does_file_version_match = True;
@@ -1222,7 +1222,7 @@ LoadGlobalState(char *filename)
 			{
 				s2++;
 			}
-			sscanf(s2, "%[^\n]", s1);
+			sscanf(s2, "%4095[^\n]", s1);
 			is_key = fxstrdup(s1);
 		}
 		else if (!strcmp(s1, "[VALUE]"))
@@ -1233,7 +1233,7 @@ LoadGlobalState(char *filename)
 			{
 				s2++;
 			}
-			sscanf(s2, "%[^\n]", s1);
+			sscanf(s2, "%4095[^\n]", s1);
 			is_value = fxstrdup(s1);
 
 			if (is_key != NULL && is_value != NULL)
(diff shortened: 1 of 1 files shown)


ThomasAdam:
Hi @Quipyowert2 

Thanks -- but I'm not fixing this with magic numbers.  If we were to do this properly, we wouldn't use `sscanf()` at all, and as I'm rewriting most of this soon anyway, I'm not keen to fix it.  It's been this way for years and never been a problem -- usually because there's a transient nature to how this file is written/read in to begin with.

Can I check -- are you a user of fvwm itself?

Quipyowert2:
I used to use fvwm2 when I had a machine running Linux but on this laptop I mainly use Windows 10 now so I don't use fvwm much anymore as a primary window manager since Windows uses its own window manager (Desktop Window Manager). When I found out about fvwm3, I wanted to help with development, so I've been running fvwm3 in Valgrind with the VcXsrv X server on WSL. I have a virtual machine too, but it runs KDE.

PR #476, opened 2021-04-16
fix shadow position bug

### LINKED ISSUE #475: Font shadow offset miscalculated
Hi ... playing around and find things ...

* Fvwm3 version: fvwm3 1.0.3 (1.0.2-109-g42414a3f)

* Linux distribution: Fedora 33

* Platform: Linux x86_64

## Expected Behaviour

Thin font shadow as with fvwm2 and fvwm3 versions older than 1. 1. 2021.

## Actual Behaviour

Font shadow has offset of 1-2 cm instead of 1 mm. (see pictures)

## Steps to Reproduce

1. Start NsCDE 1.1 or current master with fvwm3 master
2. Launch Style Manager main window and see shadows

Actual configuration is this:

It is in lines 41 and 42 of /opt/NsCDE/lib/scripts/StyleMgr but it is composed of variables and hard to read. The final result is:

`Shadow=2 0 SE:xft:DejaVu Sans:Medium:Book:size=9:antialias=0`

Changing font does not resolves this.

No

## Does Fvwm3 crash?

No

## Extra Information

Looks like this bug is introduced with d7a2eb625aac2b82ce70f493bf8c625be26aee50 or 070f35547522c30b87e922f5fe2344583ffa3880 or b92b5201eac2c1962beb367f10e757ca85e07c39 because it looks normally if I checkout some commit before d7a2eb625aac2b82ce70f493bf8c625be26aee50.

StyleMgr-Good
StyleMgr-bad

mikeandmore:
This should fix #475

### DIFF
--- a/libs/Fft.c
+++ b/libs/Fft.c
@@ -487,7 +487,8 @@ void FftDrawString(
 	if (flf->shadow_size != 0 && has_fg_pixels)
 	{
 		while (FlocaleGetShadowTextPosition(&xt, &yt, &gstp_args)) {
-			MatchFont(dpy, uf, &xt, &yt, str, len, &RenderCharFontSpec, &batch);
+			int mxt = xt, myt = yt;
+			MatchFont(dpy, uf, &mxt, &myt, str, len, &RenderCharFontSpec, &batch);
 			RenderCharFontSpec(&batch, NULL);
 		}
 	}


ThomasAdam:
@mikeandmore 

Thanks.  @NsCDE -- merged to master now, please try this.  It *looks* correct to me...

NsCDE:
> @mikeandmore
> 
> Thanks. @NsCDE -- merged to master now, please try this. It _looks_ correct to me...

Hi,

Yes, now it is correct.

Issue #393, opened 2020-12-28
FvwmCommand is missing a man page


ThomasAdam:
When `FvwmCommand` was replaced as a wrapper around `FvwmMFL` the existing man page was not included.

One needs to be written.

@NsCDE -- would you be interested?

NsCDE:
As I answered on the list, yes. Existing one has to be adapted, or new one written. But this will wait a bit because of my other issues.

ThomasAdam:
> As I answered on the list, yes. Existing one has to be adapted, or new one written. But this will wait a bit because of my other issues.

No rush at all, @NsCDE -- let me know if you need a hand.  Hope everything's going OK.

NsCDE:
Hi

Just to give icmp echo-reply ... :-)

Existing man page needs to be adapted, or new one written. But this should still wait a bit because I have to settle some private issues. I cannot plunge myself into FVWM and NsCDE this month because I don't have a good multitasking. See you!

ThomasAdam:
Hi @NsCDE 

I hope you're well.  I hate to ask, but do you think you'll have time to look at this man page soon?  The `1.0.3` release isn't far off and I'd like to include this.  If you can't do it, not a problem, as I'm sure I can put some words together.

Kindly,
Thomas

NsCDE:
Hi @ThomasAdam 

I didn't forgot this, even have a reminder. Apart from regular job, I have a load of paperwork and tasks which I must to solve IRL and I hope this will be finished soon.

I will try to put this together during this week, but I cannot promise. :-(

ThomasAdam:
Hey @NsCDE 

Not at problem -- we all have lives!  To get you started, I've created a pull-request #503 which you can use as a basis if you want it.

ThomasAdam:
Hi @NsCDE 

I've merged this for now -- if we want to do anything further for this, post the 1.0.3 release, we can do.

NsCDE:
Ok, thanks, and sorry for slowness on my side. As I said, I hope to be more on this what I like for couple of weeks.
Congrats on 1.0.3!

PR #493, opened 2021-05-24
Update manpages in default-config

somiaj:
Update the manpage menu to open the manpage for fvwm3 and remove some entries for retired modules.

### DIFF
--- a/default-config/config
+++ b/default-config/config
@@ -408,7 +408,7 @@ AddToMenu   MenuSendToPage
 # Fvwm Man Pages (Help) Menu
 DestroyMenu MenuFvwmManPages
 AddToMenu   MenuFvwmManPages "Help" Title
-+ "fvwm"                ViewManPage fvwm
++ "fvwm3"               ViewManPage fvwm3
 + "FvwmAnimate"         ViewManPage FvwmAnimate
 + "FvwmAuto"            ViewManPage FvwmAuto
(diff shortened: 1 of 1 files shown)


default-config/config
```diff
@@ -408,19 +408,17 @@ AddToMenu   MenuSendToPage
 # Fvwm Man Pages (Help) Menu
 DestroyMenu MenuFvwmManPages
 AddToMenu   MenuFvwmManPages "Help" Title
-+ "fvwm"                ViewManPage fvwm
++ "fvwm3"               ViewManPage fvwm3
 + "FvwmAnimate"         ViewManPage FvwmAnimate
 + "FvwmAuto"            ViewManPage FvwmAuto
 + "FvwmBacker"          ViewManPage FvwmBacker
 + "FvwmButtons"         ViewManPage FvwmButtons
-+ "FvwmCommand"         ViewManPage FvwmCommand
```
ThomasAdam, line 416: This line should stay, since #393 will add a man page for this.

ThomasAdam (changes requested):
Hi @somiaj 

Thanks for this.  All looks good, except the one comment to keep the `FvwmCommand` line, per #393 

If you update this, I'll merge it without any further changes.

Thanks!

default-config/config
somiaj: Thanks, I just noticed the manpage was gone, but didn't realize a new one would replace it.

Updated.

PR #495, opened 2021-05-24
Remove xpmroot link.

somiaj:
Removing very old compatibility link. xpms are not used that often anymore and fvwm-root can support .png and .svg.

### DIFF
--- a/bin/Makefile.am
+++ b/bin/Makefile.am
@@ -60,25 +60,10 @@ regenerate: \
 
 ## Create compatibility symlinks if available
 
-xpmroot_name =`echo "xpmroot" | "$(SED)" -e "${transform}"`
 fvwm_root_name =`echo "fvwm-root" | "$(SED)" -e "${transform}"`
 FvwmCommand_name =`echo "FvwmCommand" | "$(SED)" -e "${transform}"`
 
-install-exec-local:
-	@rm -f $(DESTDIR)$(bindir)/$(xpmroot_name)$(EXEEXT)
-	@$(LN_S) $(fvwm_root_name)$(EXEEXT) \
-		$(DESTDIR)$(bindir)/$(xpmroot_name)$(EXEEXT) || \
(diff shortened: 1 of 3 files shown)


bin/Makefile.am
```diff
@@ -60,25 +60,14 @@ regenerate: \
 
 ## Create compatibility symlinks if available
 
-xpmroot_name =`echo "xpmroot" | "$(SED)" -e "${transform}"`
 fvwm_root_name =`echo "fvwm-root" | "$(SED)" -e "${transform}"`
 FvwmCommand_name =`echo "FvwmCommand" | "$(SED)" -e "${transform}"`
 
 install-exec-local:
```
ThomasAdam, line 66: If `install-exec-local` is being used just for `xpmroot` root, then this line can go as well.

bin/Makefile.am
```diff
@@ -60,25 +60,14 @@ regenerate: \
...
-	@rm -f $(DESTDIR)$(bindir)/$(xpmroot_name)$(EXEEXT)
-	@$(LN_S) $(fvwm_root_name)$(EXEEXT) \
-		$(DESTDIR)$(bindir)/$(xpmroot_name)$(EXEEXT) || \
-		echo "Minor warning: $(bindir)/$(xpmroot_name)$(EXEEXT) symlink was not created"
 
 install-data-local:
```
ThomasAdam, line 68: Ditto here -- if `install-datal-local` is just for `xpmroot` then this line can go as well.

ThomasAdam (changes requested):
Hi @somiaj 

Thanks -- all looks OK.  Just querying if we can also remove a few stanzas (see inline comments).

bin/Makefile.am
somiaj: I would assume we can move the stanzas. I was unsure if we should, or to keep them there. I will remove them and test.

PR #505, opened 2021-05-29
Add DesktopConfigure to config

lgsobalvarro:
* **What does this PR do?**
Adds DesktopConfigure to config and a short explanation of the available modes.

### DIFF
--- a/default-config/config
+++ b/default-config/config
@@ -37,6 +37,17 @@ InfoStoreAdd terminal xterm
 InfoStoreAdd runcmd "dmenu_run"
 InfoStoreAdd runcmdopt "-nb '#2b4e5e'"
 
+# DesktopConfiguration sets if virtual desktops/pages are "Global" 
+# This means all screens share deskstops/pages.
+# "Per-Monitor" in witch every monitor has its own virtual desktop/pages. 
+# For this mode it's recommended that FvwmPager is configured to 
+# use the monitor configuration option, so that only windows on that 
+# specific monitor are shown. Or "Shared" witch works in the same manner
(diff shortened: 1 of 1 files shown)


default-config/config
```diff
@@ -34,6 +34,16 @@ InfoStoreAdd terminal xterm
 # This is used for "Run Command" and the Meta+Space key-binding
 InfoStoreAdd runcmd "dmenu_run -nb '#2b4e5e'"
 
+# DesktopConfiguration
+#
+# Global: This is the default setting. In Global mode all screens share desks and pages
+#
+# Per-Desktop: This mode every screen have its own deks and pages. An Pager would be required for each screen.
```
ThomasAdam, line 41: You mean `per-monitor` here.

I would say: `In this mode evey screen has its own desk and pages.    It's recommended that `FvwmPager` is configured to use the `Monitor` configuration option, so that only windows on that specific monitor are shown.

lgsobalvarro: You are correct. I'll fix it now

default-config/config
```diff
@@ -34,6 +34,16 @@ InfoStoreAdd terminal xterm
...
+#
+# Shared: This mode works in the same maner as Xmonad's or herbsfluftwm handle desks. This mode only works with 2+ screens.
```
ThomasAdam, line 43: ... `This mode works in the same manner that Xmonad or Herbsflutwm handles desks.  This mode only works with >=2 screens attached.`

ThomasAdam (changes requested):
Hi @lgsobalvarro 

Looks good, thanks!   A few very minor considerations for you...

ThomasAdam (approved):

PR #515, opened 2021-06-03
Retire FvwmTile and FvwmCascade wrappers.

somiaj:
Finally retiring the backwards compatibility wrappers for
FvwmRearrange. Use FvwmRearrange -title or FvwmRearrange -cascade
instead of the wrappers.

### DIFF
--- a/modules/FvwmRearrange/FvwmRearrange.c
+++ b/modules/FvwmRearrange/FvwmRearrange.c
@@ -87,8 +87,8 @@ int do_maximize = 0;
 int do_animate = 0;
 int do_ewmhiwa = 0;
 
-char FvwmTile;
-char FvwmCascade;
+int FvwmTile = 0;
+int FvwmCascade = 1;
 
 
 RETSIGTYPE DeadPipe(int sig)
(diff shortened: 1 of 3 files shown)


modules/FvwmRearrange/FvwmRearrange.c
```diff
@@ -402,8 +402,15 @@ void parse_args(char *s, int argc, char *argv[], int argi)
...
-      /* ignore */
+    if (!strcmp(argv[argi], "-tile")) {
+      FvwmTile = 1;
+      FvwmCascade = 0;
+      resize = 1;
+    }
+	/* Does this do anything since this is default action? */
+    else if (!strcmp(argv[argi], "-cascade")) {
+      FvwmCascade = 1;
+      FvwmTile = 0;
```
ThomasAdam, line 413: Hey @somiaj 

Hmm.  I think this should be the default (set before the check for `"-tile"`, above) -- and even if someone does pass `-cascade` through, it's a no-op as these values are already set.

somiaj: Yea, it should be the default, was more wondering if this is needed, though right now if someone did `FvwmRearrange -title -cascade` it would then set it back.

PR #516, opened 2021-06-04
Configuration tweaks

ThomasAdam:
This PR implements two changes:

1. Fixes a few spelling/typographical errors.
2. Simplifies ToggleTitle function.

### DIFF
--- a/default-config/config
+++ b/default-config/config
@@ -37,14 +37,18 @@ InfoStoreAdd terminal xterm
 InfoStoreAdd runcmd "dmenu_run"
 InfoStoreAdd runcmdopt "-nb '#2b4e5e'"
 
-# DesktopConfiguration sets if virtual desktops/pages are "Global" 
-# This means all screens share deskstops/pages.
-# "Per-Monitor" in witch every monitor has its own virtual desktop/pages. 
-# For this mode it's recommended that FvwmPager is configured to 
-# use the monitor configuration option, so that only windows on that 
-# specific monitor are shown. Or "Shared" witch works in the same manner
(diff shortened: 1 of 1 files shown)


default-config/config
```diff
@@ -37,14 +37,15 @@ InfoStoreAdd terminal xterm
...
+# DesktopConfiguration sets how fvwm handles desktops:
+#
+# * global -- the default. This means all screens show the same desktops/pages.
+# * per-monitor -- every monitor has its own virtual desktop/pages.
+#     For this mode it's recommended that FvwmPager is configured to
+#     use the monitor configuration option, so that only windows on that
+#     specific monitor are shown.
+# * shared -- which works in the same manner that Xmonad or Herbsflutwm handles
+# desks. This mode only works with >=2 monitors attached.
 #
```
somiaj, line 49: Suggested update (it is hard to describe what this does, but maybe this improve things).

```
# * shared -- the virtual desktops are shared across all monitors and each monitor shows
#     a single desktop at a time. If one monitor switches to a desktop currently being
#     viewed by another, the two monitors swap which desktop they are viewing.
#     This is similar to manner to how Xmonad or Herbsflutwm handle desks.
#     This mode only works with 2 or more monitors.
```

somiaj:
Looks good, I rebased with master, built it, and tested it out, everything works as expected. I like the new `HasTitle` conditional. I couldn't find an approve button for the review and the comment is only a suggestion (such a difficult option to articulate).

default-config/config
ThomasAdam: That's better.  Tweaked further....

Issue #409, opened 2021-01-18
Windows moving between desktops on KVM switch


hrnick:
## Upfront Information
  
$ fvwm3 --version
fvwm3 1.0.2 (released)
with support for:  ReadLine, XPM, PNG, Shape, XShm, SM, XRandR, XRender, XCursor, XFT

$ cat /etc/gentoo-release
Gentoo Base System release 2.7

$ uname -sp
Linux Intel(R) Core(TM) i7-9700K CPU @ 3.60GHz

## Actual Behaviour

I have my computer connected to a KVM switch. Whenever I switch to the other computer and back all windows end up on the current desktop and I have to manually move them back to each of their desktops respectively using MoveToDesktop.

## Expected Behaviour

I would expect the windows to stay in the desktop where they were.

## Enabling logging

I have ExplainWindowPlacement enabled but it doesn't show anything. Maybe this doesn't work in fvwm3?
BugOpts ExplainWindowPlacement On

Nothing in ~/.fvwm/fvwm3-output.log is being generated by this event.

## Steps to Reproduce

Switch to other computer connected to KVM switch and back again. This did not happen in fvwm2.

I've reduced my config to this and still have the same problem:

IgnoreModifiers            L25
DeskTopSize                1x1

Key F12         A A Menu RootMenu Root c c Nop
Key F11         A A Popup MoveToDesk
Key 1           A 4 GotoDesk 0 0
Key 2           A 4 GotoDesk 0 1
Key 3           A 4 GotoDesk 0 2
Key 4           A 4 GotoDesk 0 3

AddToFunc MoveToDesktop
+ I MoveToDesk 0 $0
+ I Prev (!Iconic, CurrentDesk, AcceptsFocus) Focus

AddToMenu MoveToDesk
+ "Desktop &1"                         MoveToDesktop 0
+ "Desktop &2"                         MoveToDesktop 1
+ "Desktop &3"                         MoveToDesktop 2
+ "Desktop &4"                         MoveToDesktop 3

AddToMenu RootMenu
+ "URxvt"               Exec urxvt
+ "Restart FVWM"                       Restart
+ "Quit FVWM"                          Quit

ThomasAdam:
Hi @hrnick 

Thanks.  I can't test this (I don't have a KVM switch -- unless you know of a means of simulating and reproducing this via KVM?), but my first thought is that it's probably RandR-related.

[...]

PR #544, opened 2021-06-18
Fix compiler warnings in modules/FvwmScript/Instructions.c

pm215:
This pull request has a couple of simple commits that fix the following compiler warnings that I see with gcc (Debian 8.3.0-6) 8.3.0:

```
Instructions.c: In function ‘FuncGetOutput’:
Instructions.c:366:7: warning: ‘strncpy’ specified bound 255 equals destination size [-Wstringop-truncation]
       strncpy(Command,cmndbuf,255);
       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
Instructions.c: In function ‘WriteToFile’:
    Instructions.c:1743:5: warning: passing argument 1 to restrict-qualified parameter aliases with argument 3 [-Wrestrict]
     sprintf(buf,"%s\n%s%d\n",buf,StrBegin,getpid());
     ^~~~~~~
Instructions.c:1744:5: warning: passing argument 1 to restrict-qualified parameter aliases with argument 3 [-Wrestrict]
     sprintf(buf,"%s%s",buf,arg[1]);
     ^~~~~~~
Instructions.c:1745:5: warning: passing argument 1 to restrict-qualified parameter aliases with argument 3 [-Wrestrict]
     sprintf(buf,"%s%s\n",buf,StrEnd);
     ^~~~~~~
```

I believe these to be genuine bugs, not just the compiler being over-enthusiastic. The caveat is that because I don't use FvwmScript myself I don't have any way of testing my fixes. I opted for the simplest changes that fix the bugs, retaining the old-school string handling rather than attempting any kind of rewrite to something less bug-prone. (My motivation for sending in this fix was mostly just as a warm-up task before I have a look into issue #409, which I've been running into.)

### DIFF
--- a/modules/FvwmScript/Instructions.c
+++ b/modules/FvwmScript/Instructions.c
@@ -56,7 +56,7 @@ extern Atom propriete;
 extern char *LastString;
 char *FvwmUserDir = NULL;
 char *BufCom;
-char Command[255]="None";
+char Command[256]="None";
 time_t TimeCom=0;
 
 /*
--- a/modules/FvwmScript/Instructions.c
+++ b/modules/FvwmScript/Instructions.c
@@ -1740,9 +1740,7 @@ static void WriteToFile (int NbArg,long *TabArg)
   }
   if (CurrPos==size)
   {
-    sprintf(buf,"%s\n%s%d\n",buf,StrBegin,getpid());
-    sprintf(buf,"%s%s",buf,arg[1]);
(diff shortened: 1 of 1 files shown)


ThomasAdam (commented):
Hi @pm215 

Thanks.  This looks OK to me, although really we should be using xasprintf() instead.  If you want to resend using that, fair enough, but it's not something I'm too fussed about either way.

ThomasAdam (approved):

PR #555, opened 2021-06-21
GNOME: remove DO_IGNORE_GNOME_HINTS

lgsobalvarro:
* **What does this PR do?**

Removes DO_IGNORE_GNOME_HINTS from fvwm/window_flags.h


ThomasAdam:
Awesome!  Thanks for this.  One small change, I would prefer the commit message read something like the following:

```
    GNOME: remove DO_IGNORE_GNOME_HINTS
    
    GnomeHints was originally added to fvwm2 when GNOME 1.x had non-standard
    window hints which fvwm2 would track for compatibility.
    
    The days of GNOME 1.x are long gone, so we can safely remove the check
    for this hint; surrounding code has already been removed.
```

PR #567, opened 2021-06-29
Cleanup SetRCDefaults()

somiaj:
This function creates a minimum config if an empty config is loaded and is the starting point for all configs. This fixes the minimum config to ensure that users who end up here have a minimum menu that can access xterm, FvwmConsole, load the default-config, or quit fvwm.

### DIFF
--- a/fvwm/fvwm3.c
+++ b/fvwm/fvwm3.c
@@ -1345,34 +1345,30 @@ static void SetRCDefaults(void)
 		{ "TitleStyle Centered -- Raised", "", "" },
 		{ "Style * Color lightgrey/dimgrey", "", "" },
 		{ "Style * HilightFore black, HilightBack grey", "", "" },
-		{ "DestroyFunc FvwmMakeMissingDesktopMenu", "", "" },
-		{ "AddToFunc   FvwmMakeMissingDesktopMenu I PipeRead 'fvwm-menu-desktop --enable-mini-icons --fvwm-icons'", "", "" },
+		{ "AddToFunc LoadDefaultConfig", "", "" },
+		{ "+ I Read "FVWM_DATADIR"/default-config/config", "", "" },
+		{ "+ I StartFunction", "", "" },
 		{
 			"AddToMenu MenuFvwmRoot \"",
 			_("Builtin Menu"),
 			"\" Title"
 		},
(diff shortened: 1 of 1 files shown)


fvwm/fvwm3.c
```diff
@@ -1403,7 +1401,8 @@ static void SetRCDefaults(void)
 		ecc.type = Restarting ? EXCT_RESTART : EXCT_INIT;
 		ecc.w.wcontext = C_ROOT;
 		exc = exc_create_context(&ecc, ECC_TYPE | ECC_WCONTEXT);
-		xasprintf(&cmd, defaults[i][0], defaults[i][1], defaults[i][2]);
+		xasprintf(&cmd, "%s%s%s", defaults[i][0], defaults[i][1],
+			defaults[i][2]);
 		execute_function(NULL, exc, cmd, 0);
```
ThomasAdam, line 1406: This change needs to be in its own commit.  I still want to see:

```
xasprintf(&cmd, "%s %s %s", defaults[i][0], defaults[i][1], defaults[i][2]);
```

So that the onus isn't on the string-handling to inject spaces -- if a string has extra spaces here, it doesn't matter.

fvwm/fvwm3.c
```diff
@@ -1338,41 +1338,39 @@ static void SetRCDefaults(void)
 		{ "SetEnv FVWM_IS_FVWM3 1", "", "" },
 		/* The below is historical -- before we had a default
 		 * configuration which defines these and more.
+		 * This is still used if an empty or custom config is loaded.
+		 * Defining a minimum default menu just in case.
```
ThomasAdam, line 1342: No need for this comment.

ThomasAdam (changes requested):

fvwm/fvwm3.c
somiaj: Extra spaces do matter because the translations are inside strings sent to fvwm for menu items and menu titles inside of quotes, so the extra spaces are not ignored. This will turn `+ "Restart fvwm" Restart` to `+ " Restart fvwm " Restart` which to me makes the menu far harder to control its look with extra spaces (why I removed it).

ThomasAdam (approved):

PR #569, opened 2021-06-29
Fvwm3 man page Cleanup

lgsobalvarro:
Fvwm has change over time. Some modules have been deprecated or replaced for one reason or another. This cleanup remove mention of those modules that are no longer available like FvwmF4 and FvwmCpp.
It also removes a small section dedicated to integration to GNOME 1. 
Some other minor edits have been made to formatting and such.


ThomasAdam:
I am not sure this is relevant any more, given that the man pages have been restructured?

PR #582, opened 2021-07-15
Allow per monitor EdgeCommand and EdgeLeaveCommand

somiaj:
Add an option `screen RANDRNAME` to `EdgeCommand` and `EdgeLeaveCommand` to allow setting the edge command for only a single monitor.

In addition make `EdgeCommand` and `EdgeLeaveCommand` honor the manual page and remove all commands if given no options are given or remove all commands for a single monitor if only given the `screen RANDRNAME` option is given.

### DIFF
--- a/doc/fvwm3/fvwm3.adoc
+++ b/doc/fvwm3/fvwm3.adoc
@@ -4909,12 +4909,14 @@ modifiers. I.e. do not ignore
 if you have no problem with it. In the _FAQ_ you can find a better
 solution of this problem.
 
-*EdgeCommand* [_direction_ [_Function_]]::
+*EdgeCommand* [screen _RANDRNAME_] [_direction_ [_Function_]]::
 	Binds a specified fvwm command _Function_ to an edge of the screen.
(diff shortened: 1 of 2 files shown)


ThomasAdam:
Hi @somiaj 

All makes sense.  Please see a commit I've pushed on top of yours on `js/per-monitor-edgecommand` to reduce code duplication.

I haven't got the time to test this, so please let me know how you get on with it, and if it's working, I'll merge the whole lot.

somiaj:
@ThomasAdam All appears to work with my tests, and the Edge{Leave}Commands can be set per monitor.

Only thing I saw is one of your comments is a bit misleading

```
	/* If action is the empty string here, then we were given
	 * "EdgeCommand" or "EdgeLeaveCommand", and no other options.  In this
	 * case, we remove all PanFrames for all monitors.
	 */
```

This doesn't remove all PanFrames, it removes the Edge{Leave}Commands, which reverts the PanFrames to their default (set via EdgeScroll), also it only removes the Edge{Leave}Commands for the stated monitor if screen is provided, not all.

PR #583, opened 2021-07-16
Remove Xinemera screen number from WindowId root.

somiaj:
With RandR there is a single root window and not one per monitor, so specifying the screen number for the root window no longer has any effect. This removes the option of specifying the screen number for the root window reducing the command to 'WindowId root [command]'.


ThomasAdam:
Hey @somiaj,

Thanks.  This falls under the work related to PR #333, so I'll include it with that when I finish it off.

PR #605, opened 2021-08-30
MoveToScreen: fix NULL-dereference

ThomasAdam:
Restore the previous behaviour of `MoveToScreen` implying the current monitor if no parameter given.


NsCDE:
That's it Thomas, it works.

PR #651, opened 2022-01-13
Fix ExitFunction

pghvlaans:
A typo in fvwm3.c (line 553) prevents ExitFunction from executing.

* **What does this PR do?**
A missing `%s` in the xasprintf call at line 553 causes ExitFunction to not execute. Adding `%s` allows ExitFunction to execute as expected.


ThomasAdam:
Thanks, @pghvlaans -- all merged.  Well-spotted!

PR #653, opened 2022-02-21
avoid sprintf(%n)

omar-polo:
Hello,

On OpenBSD the compiler warns that:

```
warning: '%n' format specifier support is deactivated and will call abort(3)
```

The `%n` conversion specifier for the printf* family of functions is widely discouraged for security reasons and the OpenBSD libc in particular will abort(3) if such specifier is found.  (scanf* %n is a different story)

### DIFF
--- a/libs/ColorUtils.c
+++ b/libs/ColorUtils.c
@@ -422,15 +422,15 @@ int pixel_to_color_string(
 
 	if (!use_hash)
 	{
-		sprintf(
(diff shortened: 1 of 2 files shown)


ThomasAdam:
Hey @omar-polo 

Awesome!  Thanks, all merged,

omar-polo:
thanks :)

PR #666, opened 2022-05-29
New default wallpapers

rasatpc:
* **What does this PR do?**
*  Adds 4 new default background wallpapers (tiled) and background icons.
* New $[FVWM_DATADIR]/default-config/images/background/bg5.png
* Default config edited accordingly.

* No screenshots.


ThomasAdam:
Hey @rasatpc 

Thanks for sending this PR through!

I think I'm going to close this though -- the entire point of the default configuration is to show intent -- which it does.  I am not sure why we'd want to include these wallpapers, as this can already be done on a per-user basis.

rasatpc:
The intent of using **tiled** wallpapers in Fvwm is unknown to most users and how to make them. Also, the current wallpapers don't encourage new users to take a second look at Fvwm when seeing the strange/ugly backgrounds. Change them or at least the bg3.png and make it the new startup default image.

ThomasAdam:
Hi @rasatpc 

I'm not unsympathetic to what you're saying, but the point of this isn't to pander to everyone's personal opinion about what should and should not be in the default configuration.

We have spent some time rewriting the default configuration -- and although it is never going to be to everyone's taste, it is at least showing the *intent* of what can be done -- and that is the important thing here, alongside the existing commented configuration, as well as the forums and the wiki.

It has always been the case that people using fvwm change things to their liking -- and the number of fvwm configuration frameworks proves that -- yours included.

I really don't want to start a precedence of changing the default configuration just so it makes existing fvwm frameworks easier -- and let's be clear here -- we're not even talking about functionality, but rather wallpapers.  Sorry, but no.  This change isn't going in.

I am not sure where you think `the current wallpapers don't encourage new users to take a second look at Fvwm when seeing the strange/ugly backgrounds.` -- can you point me to some concrete data surrounding this?

PR #682, opened 2022-08-02
Do not calculate size for missing characters.

### LINKED ISSUE #622: FvwmIconMan crashes when trying to display some utf-8 characters.
FvwmIconMan, using the default config, crashes when trying to display some utf-8 characters. For example the following character crashes FvwmIconMan

https://www.compart.com/en/unicode/U+1F479

With some further testing it is related to the font. The font from the default config is `*FvwmIconMan: Font "xft:Sans:Bold:size=8:antialias=True"`. Setting the font to `*FvwmIconMan: Font ""`, it no longer crashes and just displays a empty box for the character it cannot find. So this maybe a combination of the font and FvwmIconMan.

If possible, FvwmIconMan should better deal with some utf-8 characters in some fonts.

### LINKED ISSUE #680: Special characters leads FvwmIconMan, FvwmIdent crash.
## Upfront Information
* Fvwm3 version
`fvwm3 1.0.4 (released)
with support for:  ReadLine, XPM, PNG, SVG, Shape, XShm, SM, Bidi text, XRandR, XRender, XCursor, XFT, NLS`
`Linux 5.18.0-3-amd64 #1 SMP PREEMPT_DYNAMIC Debian 5.18.14-1 (2022-07-23) x86_64 GNU/Linux`
* Platform
`12th Gen Intel(R) Core(TM) i5-12500`
## Expected Behaviour
No unexpected crashes.
## Actual Behaviour
Special characters leads FvwmIconMan, FvwmIdent crash.
## Steps to Reproduce
1. Open any program whose title contains special characters.
e.g., use browser open following link:
[【歌枠🎤初見さん大歓迎】歌枠！ボカロ/アニソン/J-POP！【新人Vtuber/水羽そら】](https://www.youtube.com/watch?v=714hT8Qwrd0)
2. FvwmIconMan crash, FvwmIdent can't open.
I didn't try.
## Does Fvwm3 crash?
Only module, I think.
## Extra Information
* Screen capture:
螢幕截圖2022-07-30 03-32-23
螢幕截圖2022-07-30 03-33-42

topcat001:
This is my understanding of the situation: When the code in [`MatchFont`](https://github.com/fvwmorg/fvwm3/blob/4d646ca61c80545f9b0467dae68da8dbe3395f72/libs/Fft.c#L282) decides that the chosen font does not have the character in question, it constructs a fallback font from the properties of the original. Then, it reports the dimensions of the character  and also renders it using this new fallback font. Mostly this works fine. I checked this by debugging through the code character by character.

However, the only exception I found is for characters with codelen = 4. There might be other cases but I could not reproduce this in my tests with other types of chars. In this case calling `XftCharExists` on the char with the fallback font will return true and `XftTextExtents32` will return some reasonable looking values, but the character actually does not render using the fallback font when `render_char` is called next. This is the case which throws BadLength.

For `FvwmIdent`, in fact during the [`MakeList`](https://github.com/fvwmorg/fvwm3/blob/4d646ca61c80545f9b0467dae68da8dbe3395f72/modules/FvwmIdent/FvwmIdent.c#L1102) call when no rendering is done, `XftTextExtents32` is called for these problematic characters using the fallback font. This is enough to cause a BadLength error during the Xlib calls [here](https://github.com/fvwmorg/fvwm3/blob/4d646ca61c80545f9b0467dae68da8dbe3395f72/modules/FvwmIdent/FvwmIdent.c#L1251).

In my test fix I skip over these characters if a fallback is active. This avoids the Xlib BadLength error which caused the crash. Of course this test is too simplistic, but maybe we can find a better way to detect and skip such non-rendering characters. Please have a look, thanks!

```
Fixes #622, fixes #680.
```

### DIFF
--- a/libs/Fft.c
+++ b/libs/Fft.c
@@ -336,8 +336,15 @@ static void MatchFont(Display *dpy,
 
 			if (result == FcResultMatch) {
 				XftFont *fallback_font = XftFontOpenPattern(dpy, font_pattern);
-				if (fallback_font) sp.font = fallback_font;
-				else FcPatternDestroy(font_pattern);
+				/*
+				 * Hack to not use a fallback font for emoji's (codelen == 4).
+				 * FIXME: Temporary fix to avoid crashes until a proper
+				 * fix to display emoji's with a fallback font is found.
+				 */
+				if (fallback_font && codelen != 4)
+					sp.font = fallback_font;
+				else
+					FcPatternDestroy(font_pattern);
 			}
 		}
 		XGlyphInfo info;


somiaj:
I have messed with this a bit, and though it would be nice if a proper fix to correctly display emoji's via a fallback fount is created, I think it is worthwhile putting a temporary fix to keep FvwmIconMan and FvwmIdent from crashing due to emjoi characters which are becoming more and more common. Through some trial and error I have hopefully found a better place to bail out of the fallback font computation, such that a rectangular box is now shown in place of the missing character, so it is at least clear characters are not being printed. Here is my patch, with a comment to inform future users of the hack.

```diff
diff --git a/libs/Fft.c b/libs/Fft.c
index a8938dcf..7767843d 100644
--- a/libs/Fft.c
+++ b/libs/Fft.c
@@ -336,7 +336,12 @@ static void MatchFont(Display *dpy,
 
                        if (result == FcResultMatch) {
                                XftFont *fallback_font = XftFontOpenPattern(dpy, font_pattern);
-                               if (fallback_font) sp.font = fallback_font;
+                               /*
+                                * Hack to not use a fallback font for emoji's (codelen == 4).
+                                * FIXME: Temporary fix to avoid crashes until a proper
+                                * fix to display emoji's with a fallback font is found.
+                                */
+                               if (fallback_font && codelen != 4) sp.font = fallback_font;
                                else FcPatternDestroy(font_pattern);
                        }
                }
```

Since this is only an issue with fallback fonts, emoji characters should work just fine for users who use fonts with that include the characters.

It would be nice to get a real fix, but I think a temporary fix (which may last a long time) maybe useful until someone wants to dig into the issue more deeply.

topcat001:
I like this fix. Tested and it works for me, thanks!

ThomasAdam:
Hi,

Thanks  @topcat001 -- can you amend your original PR with @somiaj's solution?  Just one style nit though, the `if` statement should be on two lines, as in:

```
if (fallback_font && codelen != 4)
    sp.font = fallback_font;
else
    FcPatternDestroy(font_pattern);
```

somiaj:
Note, I just found <https://gitlab.freedesktop.org/xorg/lib/libxft/-/commit/6a08279ee80dbd1a226592d7f18c862c2e4a7d3a>, which is a bug in XFT with colored emoji's causing the BadLength error we are hacking our way around. Since this merge was only 3 months ago, maybe this fixes the issue. @topcat001 is testing and will return info if so.

somiaj:
I just built a new XFT and everything works fine, so the bug was hiding in XFT all along and was due to colored emoji's.

With a new XFT, I not only can see the emoji's in FvwmIconMan and FvwnIdent, they show up correctly in color.

This (and the various issues) should be closed as fixed in the most recent version of XFT.

There is still one issue still existing, and that is the FvwmIconMan balloon popups don't correctly use XFT fonts (FvwmPager works fine). I'll open a new issue for that.

ThomasAdam:
OK, thanks for investigating, gents!  Please close this and related issues -- if you could ensure the related issues you're closing have the label of `skip:changelog`, that'll help keep the changelog clean.

topcat001:
Ok confirmed, fvwm master indeed does not crash on latest xft on Arch. I also see the balloon issue that @somiaj mentioned for the icon manager.

somiaj:
Closing this, as the bug was fixed in XFT upstream.

PR #686, opened 2022-08-27
snapattraction: fix parsing for screen/screenall

### LINKED ISSUE #652: Snapping to screen edges doesn't work
In the current master (e327cc6) snapping to screen edges doesn't work any more (snapping to other windows still works).

Tested with a minimal `~/.fvwm/config`:

    Style * SnapAttraction 20 Windows Screen

In 42df1c8 everything snaps as expected.

ThomasAdam:
When parsing Screen and ScreenAll tokens for SnapAttraction, make sure
they're applied properly.

Fixes #652

### DIFF
--- a/fvwm/move_resize.c
+++ b/fvwm/move_resize.c
@@ -2426,7 +2426,8 @@ static void DoSnapAttract(FvwmWindow *fw, size_rect sz, position *p)
 
 	int icon_mask = IS_ICONIFIED(fw) ? SNAP_SCREEN_ICONS :
 		SNAP_SCREEN_WINDOWS;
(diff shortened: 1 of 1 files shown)


somiaj:
Overall this fixes the issue, but I think the logic is still slightly off (compared to manual page). The `Screen` by itself should only snap if something already snaps (my understanding). So `SnapAttraction 15 None Screen` shouldn't snap windows to screen, but does. Similarly `SnapAttraction 15 Icons Screen` should only snap icons to the screen, but windows are snapping too.

ThomasAdam:
Fixed already.

PR #691, opened 2022-09-04
WIP: meson.build

ThomasAdam:


g7aaa (approved):

meson.build
```diff
@@ -0,0 +1,87 @@
...
+endif
+add_project_arguments('-DVERSIONINFO=@0@'.format(version), language: 'c')
+
+# Dependencies - REQUIRED
+libevent = dependency('libevent', required: true)
+libx11 = dependency('x11', required: true)
+libxt = dependency('xt', required: true)
+xft = dependency('xft', required: true)
+xrandr = dependency('xrandr', required: true)
+xrender = dependency('xrender', required: true)
```
eli-schwartz, line 56: Note that the default value of `required` is true, so you don't need to spell it out.

meson.build
```diff
@@ -0,0 +1,87 @@
...
+fribidi = dependency('fribidi', required: false)
+ncurses = dependency('ncurses', required: false)
+libpng = dependency('libpng', required: false)
+readline = dependency('readline', required: false)
+rsvg = dependency('librsvg-2.0', required: false)
+sm = dependency('sm', required: false)
+xcursor = dependency('xcursor', required: false)
+xext = dependency('xext', required: false)
+xi = dependency('xi', required: false)
+xpm = dependency('xpm', required: false)
```
eli-schwartz, line 73: Many of these probably want to be `required: get_option('foo')` for any given feature value of foo.

This should be feature options defined in meson_options.txt, see https://mesonbuild.com/Build-options.html#features

Feature options are the equivalent of autotools `--with-foo={yes|no|check}`, although Meson uses `{enabled|disabled|auto}`.

meson.build
```diff
@@ -0,0 +1,87 @@
...
+#subdir('fvwm')
+#subdir('modules')
+
+summary({
+'libevent': libevent.found(),
+'libx11': libx11.found(),
+'libxt': libxt.found(),
+'xft': xft.found(),
+'xrandr': xrandr.found(),
+'xrender': xrender.found(),
```
eli-schwartz, line 86: You can pass the dependency objects here directly, without .found(), since Meson 0.57.0 (which you require anyway) -- this will also print boolean "YES" or "NO", but includes the version of the dependency that was found.

meson.build
ThomasAdam: Hi @eli-schwartz 

Thanks!  Fixed in the next push.

meson.build
ThomasAdam: Thanks again.  I've fixed this.

meson.build
ThomasAdam: Ditto.

PR #710, opened 2022-09-29
Don't build doc/fvwm3_manpage_source.adoc

somiaj:
This file is used as the source for doc/fvwm3.adoc and doc/fvwm3all.adoc, but isn't a standalone manpage. Don't build and install it.

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -3,7 +3,6 @@ MODULE_ADOC_SRCS = \
 	fvwm3.adoc \
 	fvwm3all.adoc \
 	fvwm3commands.adoc \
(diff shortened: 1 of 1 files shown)


ThomasAdam (changes requested):
Hi @somiaj 

This isn't correct -- I think this will need to go in `EXTRA_DIST` or something, otherwise it won't make it into the release tarball via `make distcheck`.

ThomasAdam:
@somiaj 

Can you see https://github.com/fvwmorg/fvwm3/pull/712 instead?  I think this fixes the issue.

ThomasAdam:
Fixed via #712 instead.

PR #713, opened 2022-10-01
Rename .stalonetrayrc to stalonetrayrc

wfp5p:
It's bad practice to put a dot file in a system config directory and there's no reason to make this a dot file.

I noticed this from a warning while creating an rpm for Fedora.


ThomasAdam:
Hi @wfp5p 

Thanks.  It's not bad practice at all.  But this is such a trivial change to make...

wfp5p:
Oh yeah, it's totally trivial.  I was thinking it was better to just make the change here rather than having distros carry along their own patch to make the change.

somiaj:
Might also just depend on the distro. Debian's Lintian doesn't complain about this dot file. I originally made this a dotfile so it was easy to copy into a users `$HOME` without renaming it. Though the file isn't copied over with the `FvwmScript-ConfirmCopyConfig`, and users would have to make additional changes to the configuration as well. So my original reasons aren't applicable, so renaming it is fine by me.

PR #722, opened 2022-10-16
ta/alloca

ThomasAdam:
- deprecate {Pixmap,Icon}Path
- WIP: war on alloca()

### DIFF
--- a/libs/Bindings.c
+++ b/libs/Bindings.c
@@ -827,8 +827,7 @@ KeySym FvwmStringToKeysym(Display *dpy, char *key)
 	}
 	else
 	{
-		s = alloca(strlen(key) + 1);
-		strcpy(s, key);
+		s = fxstrdup(key);
 		/* always prefer the lower case spelling if it exists */
 		*s = tolower(*s);
 		keysym = XStringToKeysym(s);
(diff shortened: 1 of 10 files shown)


ThomasAdam:
@omar-polo 

FYI -- I've started to declare war on `alloca()` -- I've not tested this, just checked it compiles.  I may well have missed some `alloca()` calls, and/pr broken functionality.

But if you wanted to use this as a starter, feel free!

libs/Bindings.c
```diff
@@ -827,8 +827,7 @@ KeySym FvwmStringToKeysym(Display *dpy, char *key)
 	}
 	else
 	{
-		s = alloca(strlen(key) + 1);
-		strcpy(s, key);
+		snprintf(s, sizeof(key), key);
```
omar-polo, line 830: This bit reads wrong: it writes to an un-initialized pointer `s`.  I believe it should be converted to either strdup or a local buffer (i.e. declaring `s` as `char s[32]` or any other sensible maximum size.)

ThomasAdam: Oops.  You're entirely correct.  Will fix.

ThomasAdam:
OK -- that's all the occurrences, except for some YACC thing in FvwmScript which I am not fussed about.  I've given this a quick test, and fvwm3 hasn't segfaulted yet...

PR #723, opened 2022-10-19
FvwmScript.c: fix wrong size of TabCom array

slazav:
I think this is an obvious error: TabCom has an explicit length in the declaration, it does not match the actual length.

FvwmScript.c:81:15: warning: type of TabCom does not match original declaration
   81 | extern void (*TabCom[25]) (int NbArg,long *TabArg);
      |               ^
Instructions.c:37:8: note: array types have different bounds
   37 | void (*TabCom[29]) (int NbArg,long *TabArg);
      |        ^
Instructions.c:37:8: note: TabCom was previously declared here


ThomasAdam:
Hi @slazav 

Yes, this looks correct.  However, your git branching needs correcting for this PR -- have a look at the following:

https://github.com/fvwmorg/fvwm3/blob/master/dev-docs/DEVELOPERS.md#branch-workflows--submitting-code-changes

PR #731, opened 2022-10-30
remove fvwm-config

ThomasAdam:
fvwm-config has often fallen behind configured features, and was
primarily used for fvwm-themes, which itself has not been updated in
many years.


somiaj:
@ThomasAdam I notice that `bin/fvwm-config.in` still exists. It isn't built, but keeping it around just in case?

PR #740, opened 2022-11-01
Ensure a default colorset is always configured.

somiaj:
Set a default colorset for active and inactive windows to ensure it is always configured if a user doesn't configure it themselves.

With removing the old color methods, no default colorset is being set (test using a blank config `fvwm3 -c /dev/null`). This ensures that a default colorset both both active and inactive windows is always configured.

### DIFF
--- a/fvwm/ConfigFvwmDefaults
+++ b/fvwm/ConfigFvwmDefaults
@@ -56,6 +56,11 @@ Style * EWMHPlacementUseDynamicWorkingArea, EWMHMaximizeUseDynamicWorkingArea
 Style fvwm_menu NoPPosition, NeverFocus, NoLenience, \
 	WindowListSkip, CirculateSkip
 
+# Default colorsets to ensure they are set.
+Colorset 0 fg black, bg grey
+Colorset 1 fg black, bg grey25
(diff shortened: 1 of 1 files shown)


ThomasAdam:
I wonder if this is a breaking change?  In the sense that now, a user is going to need to set both `Colorset` and `BorderColorset` to the same value otherwise the same colour won't be applied to the window.

somiaj:
Let me see if I understand (I didn't test this), if `BorderColorset` is not set to anything, and someone does `Colorset X`, it shouldl change both the title bar and the border? My tests shows that this isn't the case, and even if `BorderColorset` was not set, `Colorset X` would only changed the title bar, and to change the border, I have to now also use `BorderColorset`.

I think it should probably be by default that 'Colorset X' changes both the title bar and the border, unless the user explicitly sets the `BorderColorset`, and a user should probably be able to use `BorderColorset` with no options to unset the border colorset, defaulting to using the `Colorset`. So we should probably remove the setting of the `BorderColorset` from this patch and ensure that the described behavior actually works?

ThomasAdam:
Yeah... I think you've understood my ramblings.  ;)

If you're able to test your idea, please do.  If not, I'll see if I can do so over the weekend.

PR #742, opened 2022-11-03
Update Move shuffle logic to jump to next monitor

somiaj:
When using Move shuffle to move a window until it hits the next window/icon/screen edge in a given direction, windows at the EwmhBaseStruts/screen edge will move to the edge of the monitor, instead of the first object they hit.

By default the edge of the current and next monitor is computed using EwmhBaseStruts, unless the new option 'ewmhiwa' is included, in which case the EwmhBaseStruts are completely ignored and the monitor edges are used instead.

Note it is now no longer possible to shuffle to both the EwmhBaseStruts and the monitor edge. It will use one or the other but not both. If there is a panel at the EwmhBaseStruts, then with `ewmhiwa` it will still stop at the panel since it hit a window, and then will go to the screen edge after that.

### DIFF
--- a/doc/fvwm3_manpage_source.adoc
+++ b/doc/fvwm3_manpage_source.adoc
@@ -4009,14 +4009,18 @@ To move a window in a given direction until it hits another window, icon,
 or screen boundary use:
 +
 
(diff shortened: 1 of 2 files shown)


ThomasAdam:
Thanks, @somiaj -- I think this change makes sense to me.

PR #744, opened 2022-11-04
conditional: expand desk as a string

### LINKED ISSUE #741: conditions: Desk does not take a string
The `Desk` conditional parameter does not take a string as its argument and only assumes an int.  Hence, it is not possible to stipulate the following:

```
None (Desk $[monitor.$[monitor.current].desk]) Echo yep
```

This should be changed to allow for a string, expanding it via `expand_vars()` and assigning the result as an int to the relevant bitfield mask in the condition structure that `Desk` is using.

ThomasAdam:
When passing in a desk condition, such as:

    Next (Desk $[monitor.[$monitor.primary].desk]) Echo hello

This would not have worked priviously as the string was being taken
literally as an integer.

Instead, expand the string first.

Fixes #741

### DIFF
--- a/fvwm/conditional.c
+++ b/fvwm/conditional.c
@@ -43,6 +43,7 @@
 #include "decorations.h"
 #include "virtual.h"
 #include "infostore.h"
+#include "expand.h"
 
 /* ---------------------------- local definitions -------------------------- */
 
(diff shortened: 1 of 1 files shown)


somiaj:
I was looking at this, and if you go down to line 1140, `if (mask->my_flags.do_check_cond_desk)`, I noticed there is an unused if/then/else statement in the code that could be removed.

ThomasAdam:
Yeah -- it's more involved than this.  I need to treat options here as combinations, because the following won't work:

```
All (Desk 2, Screen HDMI-1) Echo "Window: $[w.name]"
```

ThomasAdam:
Fixed via #895 and also fixed the if statement logic error.

Issue #753, opened 2022-11-07
MenuStyle: deprecate {Fore,Back}Ground, Greyed, HilightBack{,Off}, ActiveFore{,Off} & MenuFace


ThomasAdam:
Per the subject those styles should be deprecated, in favour of asserting their Colorset counterparts, which already exist in code.

somiaj:
Fixed in #757

PR #757, opened 2022-11-10
MenuStyle: Remove Depreciated options and non colorset 

somiaj:
MenuStyle updates. Remove the old parser, it has been depreciated long enough. Same for the old 'FooOff' options.

Remove the ability to configure the menu using the old Foreground, Background, Greyed, and MenuFace options. This is all done via Colorsets now. This address #753 

This just removes the ability to configure or set MenuStyle via those old methods, there is still a lot of cleanup left in the code. There are multiple macros like ST_MENU_COLORS that no longer need to be used, in addition lots of conditional statements around them that can be removed, as it can now be assumed there will always be a colorset configured.

This is just a draft to start.

### DIFF
--- a/doc/fvwm3_manpage_source.adoc
+++ b/doc/fvwm3_manpage_source.adoc
@@ -2509,11 +2509,10 @@ using the *ChangeMenuStyle*. See also *DestroyMenuStyle*. When using
 monochrome color options are ignored.
 +
 _options_ is a comma separated list containing some of the keywords
-Fvwm / Mwm / Win, BorderWidth, Foreground, Background, Greyed,
-HilightBack / !HilightBack, HilightTitleBack, ActiveFore /
-!ActiveFore, MenuColorset, ActiveColorset, GreyedColorset,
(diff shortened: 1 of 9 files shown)


ThomasAdam:
Hey @somiaj

Wow. Thanks for this -- I had already started to look at this, but you've beaten me to it and done a better job of it anyway.  😀 

I think this all looks good.  One thing I was wondering though is when we're removing the commands -- can we replace those strings with the empty string? That would mean we wouldn't need to change all of the 'case' statements just because the ordering is different.

This would depend though on whether the parser is happy with the empty string or not.

somiaj:
@ThomasAdam well I took the butcher knife to the old code, everything is working here on my tests with the default-config and no-config. Though I suspect this will need a good amount of testing, by users who do more advance things with menu appearances.

somiaj:
Provided you don't find anything glaring wrong that needs fixed with this, I don't have any more thoughts on it, so you can push it to users and let the shake out the bugs.

Issue #578, opened 2021-07-12
Translucent/transparent menus


sebaro:
## Describe the new feature

* What is it which fvwm3 doesn't do at the moment, but should?

Gentoo has this patch that makes the menus transparent. It works fine and I didn't have any issue with it on FVWM2.
[fvwm-2.5.27-translucent-menus.diff.txt](https://github.com/fvwmorg/fvwm3/files/6799597/fvwm-2.5.27-translucent-menus.diff.txt)

## How might fvwm3 need to change?

* Non-technical thoughts about the impact this will have in `fvwm3`, what
  might need to change, and whether this impacts backwards-compatability.

New Colorset style options.

## Can this be achieved through scripting?

* Is a proof-of-concept available externally through scripting `fvwm3`?

No.

* Does this functionality exist elsewhere in other WMs, if so, which?

Yes.

sebaro:
It's working fine, thanks.

ExecutorElassus:
This patch, which on my gentoo applies to fvwm3-1.0.4, does not work for me for translucent menus (nor translucent colorsets applied to the buttons with 'Back' in the ButtonsBar. 

Unless there's something about the new config-file formats that I'm missing?

somiaj:
You will need to install fvwm from the master branch to get this feature, and it will be part of 1.0.6 when it is released.

somiaj:
This was adding in #760, closing.

PR #760, opened 2022-11-11
Translucent Menus

somiaj:
Update the Translucent colorset for menus patch to work with fvwm3.

Address #578 

Tested and the patch works, and has been used as an unofficial patch from various downstream versions for many years.

I did notice one issue in my testing, a menu with lots of submenus, when going down through the menu the new menu popups faster than the previous goes away, so the image grabbed to create the translucent effect will have artifacts of the previous submenu underneath that. This should probably be fixed before being merged.

I'm also unsure if this introduces any memory leaks, should be tested for that as well.


somiaj:
I have a few thoughts on this feature.

First, I don't think the Translucent option should take a color and a percent. It should just take a percent, and use the background color of the colorset. It doesn't make sense to also set the background color with this option.

Second, is this something we wish to extend to other places that use colorsets. Such as decor and modules. If we wish to extend it, it should remain a colorset option. If this is only going to be an option for menus, it should be a MenuStyle option, such as `MenuStyle * Translucent 75`. (even if we make it a MenuStyle option, it could still be extended to other places such as window decor or modules by adding a Translucent option to the appropriate place). So unless there is a strong desire to support Translucent colorsets everywhere colorsets are used, I don't think it makes sense to be a Colorset option.

ThomasAdam:
> Second, is this something we wish to extend to other places that use colorsets. Such as decor and modules. If we wish to extend it, it should remain a colorset option.

Yes, this.  As in, this should be a property of the Colorset.  I *thought* though, this was already the case, but if not, I think making it so means this will benefit other areas, as you say...

somiaj:
> I _thought_ though, this was already the case

Yes, this is currently the case. My comment was if this **only** applies to menus and not to other places that use colorsets, it might not make sense to be a colorset property.

ThomasAdam:
> Yes, this is currently the case. My comment was if this **only** applies to menus and not to other places that use colorsets, it might not make sense to be a colorset property.

Right, yes.  I agree.  Perhaps it shouldn't be a colorset property then.

somiaj:
I have updated the feature to be a `MenuStyle` option. For example to test it out just use, `MenuStyle * Translucent 75`.

The way it was implemented as a colorset was just incorrect, it shouldn't have defined its own tint to use, instead of used the already defined tints (or alphas) in the colorset. Plus now that this is a MenuStyle option, there is no expectation that this will work for modules, decor, etc. It might be possible to extend this effect to other aspects of fvwm3, but I am unsure on how to do that (and its use will be are more limited with RootTrasparent effect already supported. For cases like Panels, Pager, Iconman, they aren't often over other windows anyways). And of course if someone wants real transparency they should use a compositor.

somiaj:
I tried to make the active menu item, and title translucent, but I am unable to figure out what variables hold the correct location of the active menu item, so my attempt is slightly off, and causes a strange shadow/offset. I have put this in the code around a `#if 0` block to disable in case someone wants to take off where I left off.

If you want the active item and/or title to be translucent, you have to disable the hilight, so something like, `MenuStyle * Translucent 60, !HilightBack, !HilightTitleBack, ActiveFore` with the default-config.

sebaro:
Without a background color the menu doesn't have a good visibility on darker pars of the screen/wallpaper.
Isn't it possible to add bgcolor too, "Translucent % bgcolor"?

translucent

sebaro:
> @sebaro it uses the background color of the menu colorset, if you want to change the translucent color for the tint, change the background color of the colorset for the menu. I see no reason to also allow you to set this with the translucent option, it should just use the already defined background color of the menu.

Thanks.

somiaj:
The more I think about making it so the title and active back highlights are also translucent, the more I realize the this really isn't possible the way I'm going about it (and most likely why it was never included in the original patch). The issue is it will be grabbing a new screenshot, which includes the menu, so this will get in the way the translucent effect, since the menu is going to already have some tinting, on top of new tinting. When I get home, I am going to revert my attempt at this, and make this ready to go into the main branch (unless @ThomasAdam notices something about the code / method I'm taking that needs to be worked on).

In the future, a way to make the title and active back translucent would be to modify the translucent function to grab and store an original screenshot of the area it is using for the background of the main window. Then crop that image to the correct location for the title/menu item, and add a tint to that. This could work, but currently the original screenshot the window uses is not saved. I am not going to attempt this with this patch, and just add the feature as people are used to it work (outside of turning it into a MenuStyle). But if some future work is ever done, this idea might help them.

somiaj:
@ThomasAdam this should be ready to go, unless you notice something off about the patch that needs addressed.

ThomasAdam:
I don't see anything glaringly obvious which could hold this patch up.  Everything looks good to me.  :)

PR #768, opened 2022-11-17
Adds a subtle translucency to MenuStyle

lgsobalvarro:
Since Translucent Menus are now supported, it makes sense to showcase the feature in the default config in a non-intrusive and subtle way


somiaj:
I don't think it is worth adding translucency to the default-config, maybe a wiki page to tell people about it if needed. Should probably just close this.

lgsobalvarro:
Ok. I'm closing this then.

PR #770, opened 2022-11-21
tree/: Remove trailing whitespaces

ElyesH:
Get rid of  trailing whitespaces

Signed-off-by: Elyes Haouas <ehaouas@noos.fr>


ThomasAdam:
Hi @ElyesH 

Thanks -- but these sorts of changes aren't particularly useful.  There's all sorts of other problems across the codebase where there's a difference in style, and indentation.

Rather than doing all of this piecemeal, I'd rather see effort put into making something such as `clang-format` used -- something I started looking at.

As such I'll be closing this and your other PR.

PR #771, opened 2022-11-22
Add Style CornerLength

somiaj:
Adds a style to set the length of the corners of a window. This style
  takes a single numerical argument and sets the length of the corner
  in number of pixels beyond the edge of the boundary. If the argument
  has a '%' suffix, the number will be treated as a percent of the window
  size instead of a pixel length.

@ThomasAdam I can't seem to get the `Style * CornerLength 25%` to work consistently. Two issues, first is the flag `has_corner_percent` is just ignored unless you issue a `Style * BorderLength 5` command first, afterwards, the flag will be honored (if you issue the command again). But if you change any style such as `BorderWidth`, `HandleWidth`, `!Handles`, the flag is lost, and you need to run `Style * CornerLength 25%` again. In short I'm having trouble setting the `use_corner_percent` flag so that it correctly gets used and doesn't get reset, though it does work in some cases just fine.

### DIFF
--- a/doc/fvwm3_manpage_source.adoc
+++ b/doc/fvwm3_manpage_source.adoc
@@ -4990,10 +4990,11 @@ style.
 _options_ is a comma separated list containing one or more of the
 following keywords. Each group of style names is separated by slashes
 ('/'). The last style in these groups is the default. _BorderWidth_,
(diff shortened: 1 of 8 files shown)


somiaj:
This is just a rebase of the commit to current main. I haven't tried to fix the issues as I don't understand how the states are saved/restored.

somiaj:
Closing in favor of #928.

PR #788, opened 2022-12-29
FvwmPager: enhance rounding of windows sizes

Cigaes:
Make sure windows that are aligned on their right or bottom are also aligned in the pager. It was already guaranteed for left and top.

Instead of rounding the width and height of the window, compute and round the bottom-right coordinates and subtract to get the width and height.

before
after

### DIFF
--- a/modules/FvwmPager/x_pager.c
+++ b/modules/FvwmPager/x_pager.c
@@ -449,6 +449,7 @@ fvwmrec_to_pager(rectangle *rec, bool is_icon)
 	int m_width = monitor_get_all_widths();
 	int m_height = monitor_get_all_heights();
 	int offset_x = 0, offset_y = 0;
+	int x1, x2, y1, y2;
 
 	if (monitor_to_track != NULL) {
 		offset_x = (m_width - mon->w) * (rec->x / m_width) + mon->x;
@@ -463,16 +464,16 @@ fvwmrec_to_pager(rectangle *rec, bool is_icon)
 		scale_h = icon.height;
 	}
 
-	/* Due to rounding some windows may appear slightly off the page
-	 * edges when they are not. Should add some cleanup/snapping
-	 * code to help improve the visual effect.
-	 */
 	m_width = m_width * mon->virtual_scr.VxPages;
 	m_height = m_height * mon->virtual_scr.VyPages;
-	rec->width = (rec->width * scale_w) / m_width + 1;
-	rec->height = (rec->height * scale_h) / m_height + 1;
-	rec->x = ((rec->x - offset_x) * scale_w) / m_width;
-	rec->y = ((rec->y - offset_y) * scale_h) / m_height;
+	x1 = ((rec->x - offset_x) * scale_w) / m_width;
+	y1 = ((rec->y - offset_y) * scale_h) / m_height;
+	x2 = ((rec->x + rec->width  - offset_x) * scale_w) / m_width;
+	y2 = ((rec->y + rec->height - offset_y) * scale_h) / m_height;
+	rec->width  = x2 - x1 + 1;
+	rec->height = y2 - y1 + 1;
+	rec->x = x1;
+	rec->y = y1;
 }
 static void
 pagerrec_to_fvwm(rectangle *rec, bool is_icon)


somiaj:
This patch won't work, as now it makes windows one pixel too big. Now the bottom/right of windows are cut off if they are on the bottom/left page. The pager math rounds down for this reason due to a previous issue. The issue is the size of your pager makes this rounding visible, while in other cases it isn't. You may just want to use the patch locally.

Cigaes:
Thanks for the reply, and for not closing it before I had a chance to answer.
I can use the patch locally, of course, but would it not be better to fix it for everybody?
There is a +1 in the code that I left. Would removing it not address this? I did not want to change it without understanding where it came from, but I think it would be better to do the equivalent of adding 0.5 before rounding to get a rounding-to-nearest result.

somiaj:
It does look like that will work in my tests (though still need to test per-monitor pagers). But looking at the math, I think it will work. This should at least make it so the left end is uniformly rounded. Though I want to clean up your patch, as I don't think we need to add additional variables to the stack and can do the computation without them.

Cigaes:
Thanks for your comment. Do you want me to try to update the patch with these changes? If so, could you give me pointers to the issues that were previously solved so that I can test them too?

Regarding coding style, I added local variables because it made the formulas more readable in my opinion and this is a case where compiler optimization works well anyway, but I can do without if that is preferred.

somiaj:
Yea, there are different styles here. I have chosen to try to avoid adding unneeded variables to the code that are only used once, but added a comment to describe what the computation is doing. Anyways, I have tested the following and it seems to work (I overlooked this solution originally when I went with the scale location/size). This seems to work on per monitor pagers too. I also removed the unneeded offset variables, since they were only used once.

```diff
diff --git a/modules/FvwmPager/x_pager.c b/modules/FvwmPager/x_pager.c
index ac588ede..6ca933c3 100644
--- a/modules/FvwmPager/x_pager.c
+++ b/modules/FvwmPager/x_pager.c
@@ -448,14 +448,15 @@ fvwmrec_to_pager(rectangle *rec, bool is_icon)
 
        int m_width = monitor_get_all_widths();
        int m_height = monitor_get_all_heights();
-       int offset_x = 0, offset_y = 0;
-
        if (monitor_to_track != NULL) {
-               offset_x = (m_width - mon->w) * (rec->x / m_width) + mon->x;
-               offset_y = (m_height - mon->h) * (rec->y / m_height) + mon->y;
+               /* Offset window location based on monitor location. */
+               rec->x -= (m_width - mon->w) * (rec->x / m_width) + mon->x;
+               rec->y -= (m_height - mon->h) * (rec->y / m_height) + mon->y;
                m_width = mon->w;
                m_height = mon->h;
        }
+       m_width *= mon->virtual_scr.VxPages;
+       m_height *= mon->virtual_scr.VyPages;
 
        int scale_w = desk_w, scale_h = desk_h;
        if ( is_icon ) {
@@ -463,16 +464,16 @@ fvwmrec_to_pager(rectangle *rec, bool is_icon)
                scale_h = icon.height;
        }
 
-       /* Due to rounding some windows may appear slightly off the page
-        * edges when they are not. Should add some cleanup/snapping
-        * code to help improve the visual effect.
+       /* To deal with rounding issues, first compute the right/bottom
+        * location of the window so the rounding is consistent for windows
+        * of different sizes at the same position.
         */
-       m_width = m_width * mon->virtual_scr.VxPages;
-       m_height = m_height * mon->virtual_scr.VyPages;
-       rec->width = (rec->width * scale_w) / m_width + 1;
-       rec->height = (rec->height * scale_h) / m_height + 1;
-       rec->x = ((rec->x - offset_x) * scale_w) / m_width;
-       rec->y = ((rec->y - offset_y) * scale_h) / m_height;
+       rec->width = ((rec->x + rec->width) * scale_w) / m_width;
+       rec->height = ((rec->y + rec->height) * scale_h) / m_height;
+       rec->x = (rec->x * scale_w) / m_width;
+       rec->y = (rec->y * scale_h) / m_height;
+       rec->width -= rec->x;
+       rec->height -= rec->y;
 }
 static void
 pagerrec_to_fvwm(rectangle *rec, bool is_icon)
```

Cigaes:
Jaimos Skriletz (12022-12-30):
> Yea, there are different styles here. I have chosen to try to avoid
> adding unneeded variables to the code that are only used once, but
> added a comment to describe what the computation is doing. Anyways, I
> have tested the following and it seems to work (I overlooked this
> solution originally when I went with the scale location/size). This
> seems to work on per monitor pagers too. I also removed the unneeded
> offset variables, since they were only used once.

Thanks, I can confirm it looks well on my configuration too.

(And GitHub mangles tabs in patches :( )

Regards,

-- 
  Nicolas George

somiaj:
I think the tab clobber was copying it from my terminal, it does that, and I forgot about that.

I'll just make a new pull request.

Cigaes:
pager

Compared to fvwm2, the very small windows (FvwmButtons, music player) are a little bigger, but I can get used to it.

OTOH, I see there is a off-by-one bug on the placement of the rectangle that indicates the current position of the screen. It is a regression since version 1.0.5 as packaged by Debian.

somiaj:
I think you are running into there is a minimum window size to make it so windows don't get to small, as in the manual page.

```
       *FvwmPager: WindowMinSize n
           Specifies the minimum size as n pixels of the mini windows. This does not include the width of the border, so the
           actual minimum size is 2 * WindowBorderWidth + WindowMinSize. The default is 3.
```

Cigaes:
Indeed, WindowMinSize seems to do the trick.
As for the shift in the viewport indication, I have found that set_vp_size_and_loc() assumes the size of the pager is a multiple of the number of pages, i.e. that the size of the small screens are all the same. I am trying to transplant the formula from DrawGrid, I will post what I have when it works.

somiaj:
You also have:

```
       *FvwmPager: HideSmallWindows
           Tells FvwmPager to not show windows that are the minimum size. Useful for tiny pagers where small windows will appear
           out of place.
```

somiaj:
The pager use to be sizehints that ensured the window size of the pager was an even multiple of the number of pages/desktops being shown so that all pages were the same size. This was removed by another developer, and now the pager can be any size. Since you can control the size of the pager, best to ensure it is a multiple of the number of pages to make things work well. The issue is windows coordinates are relative to the full desktop (not a page).

Cigaes:
Thanks for the advice. `WindowMinSize 1` matches my habits very well.

As for the viewport, here is a patch that makes it use the same formula to compute the coordinates as the DrawGrid() function, it fixes the little offset I have observed.

(I have been using a buttons panel with pager of this exact size for more than 20 years now, changing it even just a little would throw me off for months.)

```diff
diff --git a/modules/FvwmPager/x_pager.c b/modules/FvwmPager/x_pager.c
index fff1887d..9853db62 100644
--- a/modules/FvwmPager/x_pager.c
+++ b/modules/FvwmPager/x_pager.c
@@ -513,12 +513,14 @@ set_vp_size_and_loc(void)
 	rectangle vp;
 	struct fpmonitor *mon = fpmonitor_this();
 
-	vp.width = desk_w / mon->virtual_scr.VxPages;
-	vp.height = desk_h / mon->virtual_scr.VyPages;
-	vp.x = (mon->virtual_scr.Vx * vp.width) /
-		mon->virtual_scr.MyDisplayWidth;
-	vp.y = (mon->virtual_scr.Vy * vp.height) /
-		mon->virtual_scr.MyDisplayHeight;
+	vp.x = (mon->virtual_scr.Vx * desk_w) /
+		mon->virtual_scr.VWidth;
+	vp.y = (mon->virtual_scr.Vy * desk_h) /
+		mon->virtual_scr.VHeight;
+	vp.width = ((mon->virtual_scr.Vx + mon->virtual_scr.VWidth / mon->virtual_scr.VxPages) * desk_w) /
+		mon->virtual_scr.VWidth - vp.x;
+	vp.height = ((mon->virtual_scr.Vy + mon->virtual_scr.VHeight / mon->virtual_scr.VyPages) * desk_h) /
+		mon->virtual_scr.VHeight - vp.y;
 	return vp;
 }
 
@@ -1587,7 +1589,7 @@ void MovePage(Bool is_new_desk)
     {
       if(i == mon->virtual_scr.CurrentDesk - desk1)
       {
-	XMoveWindow(dpy, Desks[i].CPagerWin, vp.x, vp.y);
+	XMoveResizeWindow(dpy, Desks[i].CPagerWin, vp.x, vp.y, vp.width, vp.height);
 	XLowerWindow(dpy,Desks[i].CPagerWin);
 	if (CSET_IS_TRANSPARENT(Desks[i].highcolorset))
 	{
```

Cigaes:
Note: while testing all this, I noticed two other issues :
- Running fvwm3 in Xephyr and then quitting it removes the `/tmp/fvwm_mfl.sock` from my actual session. fvwm2 used paths that depended on `$DISPLAY` and host name.
- I tried `xrandr --fb` to see if it works, and it reacted backwards.
I need to investigate and open separate issues, but I will not have the time immediately, so I just mention it here.

somiaj:
Closing in favor of #790. Thanks @Cigaes for this, and pointing out a nice way to deal with the rounding issues. Please open new issues for the last two issues you mentioned, as they are independent of this.

PR #789, opened 2023-01-01
Don't exit when enumerating screens fails

farhaven:
* **What does this PR do?**

At least on my laptop, whenever I close the lid, an XRANDR screen change event is emitted that results in 0 active monitors being returned from XRRGetMonitors. Don't exit in that case, because while it's a bit weird to have no active screens, it's not exactly untrue either.

### DIFF
--- a/libs/FScreen.c
+++ b/libs/FScreen.c
@@ -44,7 +44,7 @@ static int grav_matrix[3][3] =
 static Display *disp;
 static bool		 is_randr_present;
 
-static void		 scan_screens(Display *);
+static bool		 scan_screens(Display *);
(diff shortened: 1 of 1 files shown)


ThomasAdam:
Hi @farhaven 

Thanks for this.

I wonder if there's a neater way to solve this though.  Have a look at the `ta/monitor-count-event` branch.  Does this fix the issue for you?

ThomasAdam:
I'm going to merge that branch instead of this fix, @farhaven.  The difference here is that we don't want to change the signature of `scan_screens()` here when the logic is not about quitting `fvwm3` if no screens found, but to *only* do that if the number of screens `fvwm3` knows about is **also** 0.

Issue #104, opened 2020-05-27
reserved identifier violation


elfring:
I would like to point out that identifiers like “[`_EWMH_`](https://github.com/fvwmorg/fvwm3/blob/84499a61dfb37e6f8f2e419cc4a76266d15e8f7a/fvwm/ewmh.h#L4 "Update candidate")” and “[`_MENUS_`](https://github.com/fvwmorg/fvwm3/blob/65ba209215f9b8b27bee38c7382f1999126cc6f1/fvwm/menus.h#L3 "Another update candidate")” [do not fit](https://www.securecoding.cert.org/confluence/display/c/DCL37-C.+Do+not+declare+or+define+a+reserved+identifier#DCL37-C.Donotdeclareordefineareservedidentifier-NoncompliantCodeExample%28IncludeGuard%29 "Do not use identifiers which are reserved for the compiler implementation.") to the expected naming convention of the C language standard.
Would you like to adjust your selection for unique names?

ThomasAdam:
Hi @elfring,

Indeed, you're correct.  This is from circa 2001.

As always, patches welcome.  Bonus points from your kind self, if you use coccinelle to do it.

Always happy to be increasing the cocci scripts in `contrib/coccinelle`...

Thanks,
Thomas

elfring:
:thought_balloon: How do you think about to avoid that this software [depends on undefined behaviour](https://www.securecoding.cert.org/confluence/display/c/CC.+Undefined+Behavior#CC.UndefinedBehavior-ub_106 "Avoiding undefined behaviour?")?

PR #815, opened 2023-01-21
code: rename functions with two underscores

ThomasAdam:
The C standard discouges non-library functions beginning with a double underscore.  In fvwm's case, although some (but not all) of those were marked as static, and hence were unlikely to clash with anything in the C standard library, this change makes that consistent.


elfring:
Thanks for this source code improvement (according to [a previous bug report](https://github.com/fvwmorg/fvwm3/issues/104 "reserved identifier violation")).

PR #824, opened 2023-02-09
Revert MWM borders line width in border_fetch_mwm_layout

NsCDE:
* **What does this PR do?**

Revert MWM borders line width in border_fetch_mwm_layout

### DIFF
--- a/fvwm/borders.c
+++ b/fvwm/borders.c
@@ -1127,9 +1127,9 @@ static void border_fetch_mwm_layout(
 	ret_size_descr->w_hiout = 2;
 	ret_size_descr->w_trout = 0;
 	ret_size_descr->w_trin = 0;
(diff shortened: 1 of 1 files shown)


NsCDE:
Visual difference ...
badmwmborder
goodmwmborder

ThomasAdam:
This change is not correct with respect to HiDPI displays, where accepting this patch is now a regression.

NsCDE:
I had Philipps and now I have new monitor, some Dell 29" which is 2560x1440 - my sweet spot resolution. I don't know if it is or is not HiDPI. It looks like I will have to go to living room with HDMI cable and laptop and try this on 4K TV and max resolution. :-) Did it once.

Altrough, I'm not sure how to tell if something is or is not "HiDPI". I must find that before ...

PR #860, opened 2023-06-25
we don't need iconv under XFT

mikeandmore:
We shouldn't need to iconv to ISO-8859-1 under XFT. We need this before to deal with xfonts, xfonts encoding uses ISO-8859-1 by default (or any custom locale before the UTF-8 age).


ThomasAdam:
Fixed via #862 -- I just reworded the commit message.

PR #914, opened 2023-10-17
Fix implicit function declaration in configure

Kangie:
* **What does this PR do?**

Add missing `unistd.h` include for `close` and `unlink`


Kangie:
See https://bugs.gentoo.org/901275

PR #916, opened 2023-10-20
EdgeScroll: implement per monitor

ThomasAdam:
Bring the EdgeScroll command inline with other command, such as
EwmhBaseStruts whereby the dimensions can be specified per-monitor, such
as:

    EdgeScroll 0 0
    EdgeScroll screen DP-1 100 100

### DIFF
--- a/doc/fvwm3_manpage_source.adoc
+++ b/doc/fvwm3_manpage_source.adoc
@@ -7390,9 +7390,13 @@ Style * EdgeMoveResistance moving screen-scrolling
 +
 Fvwm does this substitution automatically and prints a warning.
 
-*EdgeScroll* _horizontal_[p] _vertical_[p] [wrap | wrapx | wrapy]::
+*EdgeScroll* [_screen RANDRNAME_] _horizontal_[p] _vertical_[p] [wrap | wrapx | wrapy]::
 	Specifies the percentage of a page to scroll when the cursor hits the
-	edge of a page. A trailing '_p_' changes the interpretation to mean
+	edge of a page.
+	The optional '_screen RANDRNAME_' specifies the RandR monitor which
+	this setting should apply to, ignoring all other monitors.  Without
+	this option, it applies the value to all monitors.
+	A trailing '_p_' changes the interpretation to mean
 	pixels. If you do not want any paging or scrolling when you hit the
 	edge of a page include
 +
@@ -7426,10 +7430,14 @@ EdgeScroll 100000 100000
 is used fvwm scrolls by whole pages, wrapping around at the edge of
 the desktop.
 
-*EdgeThickness* 0 | 1 | 2::
+*EdgeThickness* [_screen RANDRNAME_] 0 | 1 | 2::
 	This is the width or height of the invisible window that fvwm creates
 	on the edges of the screen that are used for the edge scrolling
 	feature.
++
+The optional '_screen RANDRNAME_' specifies the RandR monitor which
+this setting should apply to, ignoring all other monitors.  Without
+this option, it applies the value to all monitors.
 +
 In order to enable page scrolling via the mouse, four windows called
 the "pan frames" are placed at the very edge of the screen. This is
--- a/fvwm/events.c
+++ b/fvwm/events.c
@@ -4186,7 +4186,7 @@ void dispatch_event(XEvent *e)
 			XRRUpdateConfiguration(e);
 			monitor_update_ewmh();
 			monitor_emit_broadcast();
-			initPanFrames();
+			initPanFrames(NULL);
 			break;
 		}
 	}
--- a/fvwm/virtual.c
+++ b/fvwm/virtual.c
@@ -733,6 +733,8 @@ int HandlePaging(
 	int mwidth, mheight;
 	int edge_thickness = m->virtual_scr.edge_thickness;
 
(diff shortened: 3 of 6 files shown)


somiaj:
Tested this out and it works as described. I am able to control `EdgeScroll` on a per monitor basis.

It would also be nice to have a way to turn panframes on/off on a per-monitor basis as well.  I use `EdgeThickness 0` and `EdgeThickness 2` to turn them off/on.  One thing I noticed is that just setting `EdgeScroll screen DP-1 0 0` disables EdgeScroll on that monitor, but I can still move windows over the window boundary, as opposed to `EdgeThickness 0` disables the panframes and I cannot scroll or move windows over page boundaries.

ThomasAdam:
> One thing I noticed is that just setting `EdgeScroll screen DP-1 0 0` disables EdgeScroll on that monitor, but I can still move windows over the window boundary, as opposed to `EdgeThickness 0` disables the panframes and I cannot scroll or move windows over page boundaries.

Yeah -- OK, that's a good point.  Because the panframes still exist with setting `EdgeScroll 0 0`, I'll have to make `EdgeThickness` also per-monitor aware.

I'll add that in with this PR.

Note that the semantics of all of this is getting weird, with `EdgeResistance`, and `EdgeMoveDelay` -- which is a Style option.   But never mind, that's not something I want to address.

So I think what I'm suggesting is to achieve what you want, it is going to have to be a combination of a few different commands, but you can always put them in a function, perhaps passing in the monitor name to toggle these things on/off.

ThomasAdam:
Right, this PR now allows for:

```
EdgeThickness screen DP-1 0
EdgeScroll screen DP-1 0 0
```

Which will completely disable the ability to switch pages and drag windows on the specified monitor.  Not specifying the `screen XXX` option to those commands still works in the global way it currently does.

Please do test it, and let me know how you get on.

somiaj:
I notice a few issues with this, first `EdgeThickness` is broken if no screen is supplied, and second `EdgeThickness screeen DP-1` only seems to apply to my primary monitor, no matter what identifier I use for the screen.

somiaj:
Actually `EdgeThickness` without a screen setting is working, it is just only applying to the primary monitor, and not my secondary one.

ThomasAdam:
> Actually `EdgeThickness` without a screen setting is working, it is just only applying to the primary monitor, and not my secondary one.

Huh.  Yeah, OK.  Fix incoming...

ThomasAdam:
> Huh. Yeah, OK. Fix incoming...

Try that, when you get a mo...

somiaj:
There is still some oddness going on, but harder to describe. In short it starts out working just fine, but after toggling `EdgeThickness 0` and `EdgeThickness 2` a few times, the pan frames seem to just vanish and all scrolling stops even with `EdgeThickness 2` set, and sometimes only one monitor is affected, sometimes only one panframe (so I can scroll up but not down).

ThomasAdam:
> There is still some oddness going on, but harder to describe. In short it starts out working just fine, but after toggling `EdgeThickness 0` and `EdgeThickness 2` a few times, the pan frames seem to just vanish and all scrolling stops even with `EdgeThickness 2` set, and sometimes only one monitor is affected, sometimes only one panframe (so I can scroll up but not down).

Odd -- thanks for checking.  I'll add some logging to see where it's getting stuck.

somiaj:
Some more tests, I run `EdgeThickness screen DP-4 0`, then `EdgeThickness screen DP-4 2`, then `EdgeThickness screen DP-4 0`, at this moment the bottom panframe is removed but the top panframe is still there. I then scroll up using the panframe, but at that point all panframes are gone and I can't get them back with `EdgeThickness screen DP-4 2`.

somiaj:
For what it is worth, I have two monitors with same resolution positioned left/right and the following setup, so I can scroll up and down on each monitor.

```
# Desktop Setup
DesktopConfiguration per-monitor
DesktopSize 1x3
```

ThomasAdam:
Thanks, @somiaj 

I think this might have been due to a former check which tried to be clever about how to map/unmap the pan frames.  Now that Edge{SCroll,Thickness} can always change the state of these, it's not too expensive to recreate them on-the-fly.

So I've removed this check, just to see if that makes a difference.  From what I can tell here, it seems to.

somiaj:
My initial tests seem to confirm it is now working, I will use this for a few days to see if I have any more issues.

somiaj:
Again so far, so good. Now scroll lock can toggle scrolling on a per monitor basis.

```
DestroyFunc ToggleScroll
AddToFunc ToggleScroll
+ I Test (EnvMatch infostore.DisableScroll$[monitor.current] True) SetEdgeThickness 2 False
+ I TestRc (NoMatch) SetEdgeThickness 0 True

DestroyFunc SetEdgeThickness
AddToFunc SetEdgeThickness
+ I EdgeThickness screen $[monitor.current] $0
+ I InfoStoreAdd DisableScroll$[monitor.current] $1

Key Pause A A ToggleScroll
```

somiaj (approved):
Been using this for a little bit now, and it seems to work. I say it is good to go (though maybe rename the WIP in the commit name).

ThomasAdam:
Thanks, @somiaj -- merging this now to `main` -- let me know if you experience any other issues.

PR #919, opened 2023-10-23
ta/monitor ordering

ThomasAdam:
- WIP
- WIP: Monitor ordering


somiaj:
The order only seems to be set at the initialization stage, if I run `xrandr --output DP-2 --left-of DP-4` to change my monitor order, the variables are not updated, and now 1 is my left monitor and 0 is my right monitor. But if I then restart fvwm3, the order is updated.

Now that I see what you are doing with this, I think my lexographic order approach would be worthwhile. If the monitors are all in the same line, you'll get the same result here, but there will also be a predictable order for other setups as well. (well I'm assuming in a line means the top coordinate is all the same, with different sized monitors lined up along the bottom, which I think is a very odd setup, it would be different).

ThomasAdam:
Hey @somiaj 

Yeah -- there's definitely some rough edges which I've yet to deal with, such as recreating the ordering should the layout change while FVWM is running.

I completely agree about your lexicographical approach -- I'm just not too sure how best to implement it.

somiaj:
@ThomasAdam point me to which part of the code assigns the order, I might be able to come up with something, should be just a simple sort, and since the number of monitors will always be small, a standard sort algorithm should do the trick.

somiaj:
@ThomasAdam It seems like `qsort` in `<stdlib.h>` will do the job. Here is some pseudo code that I hope is enough for you to understand and maybe implement.

First you need a compare function, such as:

```c
int mon_compare(const void *m1, const void *m2)
{
	const struct monitor *mon1 = m1;
	const struct monitor *mon2 = m2;

	if (mon1->si->y == mon2->si->y) {
		return mon2->si->x - mon1->si->x;
	} else {
		return mon2->si->y - mon1->si->y;
	}
}
```

Then just put all of the monitors into an array, `mon_array`, and call `qsort`:

```c
qsort(mon_array, number_of_monitors, sizeof(monitor), mon_compare);
```

The array will be sorted, then just number the monitors based on their position in the array:

```c
for (int i = 0; i < number_of_monitors; i++) {
	mon_array[i]->si->number = i;
}
```

My c knowledge is still very basic, and just following examples, so I'm sure this will need to be adjusted.

libs/FScreen.c
```diff
@@ -432,17 +437,69 @@ monitor_output_change(Display *dpy, XRRScreenChangeNotifyEvent *e)
...
+		return -1;
+	else if (a->si->y > b->si->y)
+		return 1;
+	else {
+		if (a->si->x < b->si->x)
+			return -1;
+		else if (a->si->x > b->si->x)
+			return 1;
+	}
+	return 0;
```
somiaj, line 496: Optimization?

```
return (a->si->y == b->si->y) ? a->si->x - b->si->x : a->si->y - b->si->y;
```

ThomasAdam: That'll do nicely -- thanks!

somiaj:
My quick tests (with only two monitors) seem to show it works. I even tested my optimization suggestion and I get the exact same result. It still doesn't update the ordering on monitor changes, I have to restart fvwm to get the ordering to update.

ThomasAdam:
> It still doesn't update the ordering on monitor changes, I have to restart fvwm to get the ordering to update.

Really?  I'm moving the ordering of screens around using xrandr, and they update just fine, without restarting fvwm.

What am I missing?

somiaj (approved):
Seems to work now!!

PR #920, opened 2023-10-24
Add configure option to generate html man pages.

somiaj:
`./configure --enable-htmldoc` will now use asciidoctor to build and then install .html versions of the manual pages (with a table of contents) in `$FVWM_DATADIR/htmldoc`.


configure.ac
```diff
@@ -321,6 +321,35 @@ if test ! x"$with_mandoc" = xno; then
...
+)
+
+if test ! x"$with_htmldoc" = xno; then
+	if test x"$ASCIIDOC" = x ; then
+		with_htmldoc="no"
+		problem_htmldoc=": No asciidoctor found in PATH"
+	elif test x"$SED" = x ; then
+		with_htmldoc="no"
+		problem_htmldoc=": No sed found in PATH"
+	elif test x"$HAVE_PERL_FOR_DOCS" = x0 ; then
```
ThomasAdam: Why is this check necessary?

ThomasAdam:
Hi @somiaj 

Thanks for this -- things looks OK to me.  Just not sure about the $SED check.  There's perhaps one or two style cleanups I'd recommend, but nothing blocking this to be merged, IMO.

If you can just clarify the $SED point, I'll probably merge it later.

configure.ac
somiaj: I just copied the manpage checks, but the second commit removes these perl checks, since the POD is no longer being used.

somiaj:
Oh you were talking about the SED check. I removed them for both htmldoc and manpages, I don't see sed being used for either.

ThomasAdam:
Thanks!  All merged.

PR #940, opened 2023-12-23
randr: allow geometry strings to include @n

ThomasAdam:
When geometry string contain @n, where n is a number from 0..INT_MAX, treat that number as a monitor's assigned number to use.

This is similar to how fvwm used to handle these geometry strings with Xinerama, except that the designation of monitor numbers is now derived via a deterministic algorithm.  See 'RANDR SUPPORT' in "man fvwm3all"


ThomasAdam:
fvwm-geometries

Seems to work OK for me.  Testing welcome.

PR #944, opened 2024-01-10
WIP Meson Build System Switch

### LINKED ISSUE #89: Autotools: migrate to Meson
FVWM's build system is old, and `configure.ac` has a lot of legacy/dead code which is no longer required.

Added to which, some of the m4 macros in use could probably be replaced with newer constructs in autotools since the time they were written some years ago.

It would be nice to audit this.

For example:

* `imagepath` can be set here, but is configurable via runtime, so should be removed.  It's already set via `fvwm/fvwm3.c`...

polarbub:
WIP to switch the build system to meson. 
Fixes #89


ThomasAdam:
Closing this in favour of #996

PR #964, opened 2024-03-18
FvwmPager: remove list_focus handling

ThomasAdam:
This was only ever used internally by FvwmPager to keep track of which
monitor is currently focused.  This is no longer required now that this
information is used directly from libfvwm3.


somiaj:
Why not just remove the whole function (vs it do nothing)?

ThomasAdam:
I'm about to add different functionality to that, but it might not be in the 1.1.0 release.

PR #972, opened 2024-03-23
Add per-monitor support to FvwmPager

somiaj:
This adds per-monitor support to FvwmPager, in that if a pager isn't tracking a single monitor, the pager (and icon pager) will now highlight just the portion of the virtual desktop being viewed by each monitor. When using left click to move to a location in the pager, this will move the appropriate monitor to view the area, allowing one to control the monitors locations independently in a single page. The pager can also be used to move windows between monitors with middle click.

More testing of this is needed. It is mostly working in my tests. There are some known issues.

+ ~~Adding/removing monitors will cause FvwmPager to crash or behave badly, the pager needs to be reloaded to initialize the monitors correctly.~~
+ ~~Right click scroll doesn't correctly work (only scrolls the current monitor).~~ ~~(Still a small bug with scrolling the non current monitor on a different desk).~~
+ ~~When using the pager to move windows between different desks and different monitors, the window may not be put on the correct desk or could also loose sync with fvwm and where you view the window is not its real location. Moving windows on the same desk between monitors or different desks same monitor seems to work.~~
+ ~~Sticky windows are considered on the desk of the current monitor, not necessarily the monitor they are being viewed on. This will cause sticky windows to appear on the wrong desk when viewing multiple desktops.~~

Please test this. I could really use testing `DesktopConfiguration shared`, and testing on setups with different sized monitors.


somiaj:
Almost everything is working fine. Others need to test this a bit more.

The only known issue is moving windows between different desks and different monitors causes fvwm to loose track where the window / monitor it is on (moving windows between monitors on the same desk or different desks/same monitor work as expected). I say this is okay for now (I think this has always been an issue and not new with these changes which I think have drastically improved the stability and reliability of the pager).

ThomasAdam:
Closing this in favour of #978 which is the same code as this, but rebased and cleaned up a little to remove WIP patches, etc.

PR #975, opened 2024-03-24
FvwmPager: Allow for some dynamic config loading

ThomasAdam:
This makes FvwmPager listen to 'SendToModule' commands such that the
following options can be configured on a running FvwmPager instance:
    
* DeskLabels
* NoDeskLabels
* Monitor RANDRNAME
* MonitorLabels
* NoMonitorLabels
    
For example:
    
   SendToModule FvwmPager Monitor $[monitor.1.name]
    
Will force an already-running FvwmPager instance to only track that monitor.

### DIFF
--- a/doc/FvwmPager.adoc
+++ b/doc/FvwmPager.adoc
@@ -371,6 +371,27 @@ done about this - except not using SloppyFocus in the pager.
   is especially meaningful when the _DesktopConfiguration_ command is
   set to _shared_.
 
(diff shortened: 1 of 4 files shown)


ThomasAdam:
Before merging this, we should also consider which other settings we want to be configurable:

* MonitorLabels
* DeskLabels

etc, etc.

ThomasAdam:
Closing.  To be replaced with a different PR.

PR #981, opened 2024-03-30
FvwmMFL: add DISPLAY to FVWMML_SOCKET

ThomasAdam:
This adds the DISPLAY information to the given FVWMML_SOCKET, and
updates FVWM's environment with that new information.  This is a hackish
way to make FVWMMFL sockets unique per-display.

This needs a rethink.


ThomasAdam:
This isn't being implemented like this.

PR #987, opened 2024-04-02
Use AX_COMPARE_VERSION macro for golang version checks

Kangie:
The previous approach fragile, required manual updates, and results in unexpected outcomes when the user has updated golang.

* **What does this PR do?**

We stop manually whitelisting golang versions in favour of checking against a minimum version.

I know that meson is coming (yay!), but this keeps biting us in Gentoo. :)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -80,23 +80,13 @@ AC_ARG_ENABLE(golang,
 
 if test ! x"$with_golang" = xno; then
 	AC_CHECK_PROGS(GO, go)
-	if test -n "$GO" ; then
-	    GOVERSIONOPTION=version
-	    go_version=$($GO $GOVERSIONOPTION | sed -e 's/go version go//' | \
-		sed -e 's/ .*$//')
-	    AC_MSG_CHECKING([whether go version is >= 1.14.x ($go_version)])
-	    case "$go_version" in
-	     1.14*|1.15*|1.16*|1.17*|1.18*|1.19*|1.20*|1.21*|1.22*|1.23*)
-	      AC_MSG_RESULT([yes - version is: $go_version])
-	      with_golang="yes"
-	      GO=
-	      ;;
-	    *)
-	      AC_MSG_RESULT([no - version is: $go_version])
-	      with_golang="no"
-	      problem_golang=": version of go ($go_version) <= 1.14.x"
-	      ;;
-	    esac
+	if test -n "$GO"; then
+		GOVERSIONOPTION=version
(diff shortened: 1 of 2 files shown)


ThomasAdam:
@Kangie 

Thanks -- I was hoping there would be an easier way.

But you've got syntax issues with this change.  CI isn't passing.  Please fix.

Kangie:
Ah. We can either ship m4/ax-compare-version.m4 or add autotools-archive to the fvwm3-build container.

ThomasAdam:
@Kangie 

Thanks.  But, please only make the necessary changes required to introduce `AX_COOMPARE_VERSION` -- I am not interested in any other autotools/automake/configure changes at this time.

ThomasAdam:
> Do you have a preference around updating the build image or shipping the m4 macro in-repo?

Add it to `acinclude.m4`.

Kangie:
Dropped the offending commits.

Please be aware of https://bugs.gentoo.org/919163

Is it worth hacking away at the meson build path instead?

configure.ac
```diff
@@ -80,23 +80,13 @@ AC_ARG_ENABLE(golang,
...
-	      AC_MSG_RESULT([no - version is: $go_version])
-	      with_golang="no"
-	      problem_golang=": version of go ($go_version) <= 1.14.x"
-	      ;;
-	    esac
+	if test -n "$GO"; then
+		GOVERSIONOPTION=version
+		go_version=$($GO $GOVERSIONOPTION | sed -e 's/go version go//' | sed -e 's/ .*$//')
+		AX_COMPARE_VERSION([$go_version], [ge], ["1.14.0"],
+							[with_golang="yes ($go_version)"; GO=],
```
ThomasAdam, line 87: Careful here -- this change isn't quite right.  We need to set `with_golang="yes"`, and nothing else.  This is because of this check on line 92:

```
AM_CONDITIONAL([FVWM_BUILD_GOLANG], [test x"$with_golang" = xyes])
```

ThomasAdam (changes requested):
Hi @Kangie 

Thanks -- one slight change, please.  I've left an inline comment.

Kangie:
That should take care of it. I still think that the autotools tidyup I did will make maintaining the code and build system easier going forward, regardless of the system used (i.e. less complexity for meson). Would you consider a separate PR?

ThomasAdam:
Hi @Kangie 

Thanks. Will merge this.

I'm aware of some of the configure problems, and have some patches myself to address a few of them. For example, the select problem.

I'd much rather effort was put into the meson build...

PR #990, opened 2024-04-02
Don't use a declaration state in for loops

derekschrock:
This breaks builds with older gcc ex: EL7 gcc 4.8.5

* **What does this PR do?**

Fixes builds on EL7 gcc 4.8.5


ThomasAdam:
@derekschrock 

OK, but note that I will make absolutely no claims in the future about very old systems or toolchains being supported. 

Where >=C99 as a standard makes sense to use, that's exactly what I will do. This is just a small example of that.

PR #1004, opened 2024-04-10
Add a utils script to create manpage html output for www.fvwm.org

somiaj:
This script when run in the top level of the fvwm3 source will create html output designed for www.fvwm.org. The majority is done using asciidoctor in embedded mode to no longer include headers/footers. There is some additional edits done with sed to make the website format better without messing with its css too much.

Also found a minor bug in the manual page generation when testing this out, throwing it in here.


somiaj:
Closing. Move script to the website source.

PR #1013, opened 2024-04-16
FvwmPager - DeskStyles

somiaj:
DeskStyles provide a more standardized way to apply settings to each desktop individually (which can be used both when viewing multiple desks or just viewing only the current active desk). This restructures how to configure setting mostly colorsets settings on a per desk basis. With this the colors of the desktop, active monitors, windows, focus windows, and balloon window can all be configured for individual desktops. This also uses a standardized format to configure all the different colorsets or colors.

```
# Set a colorset for all desks, both syntax's do the same thing.
*FvwmPager: Colorset 12
*FvwmPager: HilightColorset * 12
```

To set a configuration for a specific desk, let's say desk 2, provide the desk number as the first parameter.

```
# Set colorset for desk 2.
*FvwmPager: Colorset 2 14
*FvwmPager: HilightColorset 2 15
```

The FvwmPager manual page is updated to explain the way to many settings that can be configured. It was also restructured to give some format to the large number of customizable options.

myscrot

myscrot

### DIFF
--- a/doc/FvwmPager.adoc
+++ b/doc/FvwmPager.adoc
@@ -214,11 +214,19 @@ size may be slightly different than the specified size.
 For backwards compatibility the option, '*FvwmPager: DeskColor [desk] color',
 is an alias for this option and works the same.
 
-*FvwmPager: Hilight [desk] color::
-  Sets the background color for both the location of any active monitors (if
-  _DeskHilight_ is set) and the current desk label and/or monitor label if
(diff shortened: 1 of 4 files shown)


somiaj:
The colors even work when only showing the current desk and `CurrentDeskPerMonitor`, showing the color of the desktop the respective windows are in.

myscrot

PR #1017, opened 2024-04-24
FvwmPager: Make mouse bindings configurable.

somiaj:
Instead of the mouse bindings being hard coded (mouse 1 changes desk/page, 2 focus/moves windows, and 3 scroll), this adds support to configure what the bindings do and supports the primary mouse buttons 1-5 only (no extended mouse buttons).

The bindings can be used to disable or move which button any of the default bindings are on, or add custom commands to send to fvwm/windows. One example is one could make the mouse wheel (buttons 4 and 5) change the current desk, so using the mouse wheel on the pager will cycle through desktops (could be useful with the always show current desk pager). You can also make bindings that act on the window you click on, so you could maximize, iconify, shade, or run any custom fvwm function on the window clicked.

The bindings are configured using the following:

```
*FvwmPager: Mouse <N> ChangePage
*FvwmPager: Mouse <N> MoveWindow [command]
*FvwmPager: Mouse <N> WindowCmd command
*FvwmPager: Mouse <N> Scroll
*FvwmPager: Mouse <N> Cmd command
*FvwmPager: Mouse <N> Nop
```

This is built on top of #1015, so that will need to be merged first.

### DIFF
--- a/doc/FvwmPager.adoc
+++ b/doc/FvwmPager.adoc
@@ -45,37 +45,33 @@ window. If no _alias_ is given, the default "FvwmPager" alias is used. See the
 FvwmPager displays a miniature view of the fvwm virtual desktop(s) showing
 the position of all windows and pages within each visible desktop. If
 _DeskHilight_ is set, the location of each monitor within the
-desktop is also shown. The pager can be used as a quick reference of the
-location of windows and monitors, to change the current page/desk, and to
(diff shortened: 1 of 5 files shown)


ThomasAdam:
Hi,

This looks fine to me.  I've only focused on the commit "add mouse binding configuration" as the other commits have already been merged.

Very minor typo, but might as will fix as this PR will need rebasing anyway, before it can be merged.

PR #1030, opened 2024-05-29
FvwmButtons: stop listening for incoming TCP connections

### LINKED ISSUE #1029: FvwmButtons listens for incoming TCP connections
Hello,

I cannot find a dedicated security contact, so I am posting this here.

FvwmButtons listens for incoming TCP connections from any host, which is a major security risk.
```
# netstat -ltpv 
Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name    
tcp        0      0 0.0.0.0:38221           0.0.0.0:*               LISTEN      15701/FvwmButtons   
tcp6       0      0 [::]:42179              [::]:*                  LISTEN      15701/FvwmButtons   
```
Running on `Linux x86_64`.

As far as I can see, the port is opened here:
https://github.com/fvwmorg/fvwm3/blob/5d6c0457f6385bc9537b62ea9fe5f4a83ee89a33/libs/fsm.c#L1046

Others seem to be mitigating the problem by limiting communication to UNIX sockets via calling [`_IceTransNoListen ("tcp")`](https://github.com/search?q=_IceTransNoListen&type=code)

Can FVWM do the same?

Thanks,
V.

vladrich:
Fixes #1029

### DIFF
--- a/libs/fsm.c
+++ b/libs/fsm.c
@@ -1043,6 +1043,12 @@ int fsm_init(char *module)
 		return 0;
 	}
 
+
+#ifdef HAVE__ICETRANSNOLISTEN
+	_IceTransNoListen("tcp");
+#endif
+
+
 	if (!FIceListenForConnections (
 		    &numTransports, &slistenObjs, 256, errormsg))
 	{
--- a/configure.ac
+++ b/configure.ac
@@ -377,6 +377,10 @@ if test ! x"$with_sm" = xno; then
   AC_CHECK_LIB(SM, SmcOpenConnection, with_sm=yes; AC_DEFINE(SESSION),
     with_sm=no; problem_sm=": Failed to detect libSM",
     [$X_LIBS $X_PRE_LIBS -lX11 $X_EXTRA_LIBS])
+  ac_LIBS="$LIBS"
+  LIBS="$LIBS -lICE"
+  AC_CHECK_FUNCS([_IceTransNoListen])
+  LIBS="$ac_LIBS"
 fi
 dnl AC_SUBST(sm_LIBS)
 dnl AC_SUBST(sm_CFLAGS)


libs/fsm.c
```diff
@@ -1043,6 +1043,18 @@ int fsm_init(char *module)
...
+	_IceTransNoListen("tcp");
+#else
+	fvwm_debug(__func__, "[%s][fsm_init] "
+		   "Security risk: _IceTransNoListen is not available. "
+		   "Cannot stop listening for TCP connections.\n"
+		   "A firewall is strongly recommended. "
+		   "See https://github.com/fvwmorg/fvwm3/issues/1029\n", module_name);
+#endif
+
+
```
ThomasAdam, line 1057: Absolutely not!

There is no need for this to be logged like this.

First of all, I don't want to log sensationalist messages like this.

Second of all, there's absolutely no security risk at all with FvwmButtons listening on TCP.  FVWM already has an open socket to send commands to (which also includes executing arbitrary commands), so there's no explicit exploit here.

Please remove this entire block, and stop inciting unnecessary FUD.

ThomasAdam:
Thanks, @vladrich 

There's no way I'm merging this as-is.  Comments and corrections given to the PR.

libs/fsm.c
```diff
@@ -1043,6 +1043,16 @@ int fsm_init(char *module)
...
+#ifdef HAVE__ICETRANSNOLISTEN
+	_IceTransNoListen("tcp");
+#else
+	fvwm_debug(__func__, "[%s][fsm_init] "
+		   "_IceTransNoListen is not available. "
+		   "Cannot stop listening for TCP connections.\n", module_name);
+#endif
+
+
```
ThomasAdam, line 1055: Skip the else block entirely, please.

vladrich: I have removed the output completely now. I do want to point out though that without `_IceTransNoListen`, FVWM is accepting **remote** connections.  Anyone with a network route to the machine can, in the very least, crash FvwmButtons. Unless I am missing something here, this does look like a security risk to me.

ThomasAdam: Actually, no. It's not a big security risk. 

I will merge this.

PR #1035, opened 2024-06-10
FvwmCommand: use FVWMMFL_SOCKET when it's present

### LINKED ISSUE #1034: FvwmCommand doesn't find socket
After commit da4387bd84c6dcf7e4483d545c48610e5420992b, I started using FVWMMFL_SOCKET_PATH to put sockets into $XDG_RUNTIME_DIR/fvwm (which is the right place for multi-user system)

FvwmMFL (from the default config) creates a socket and provides FVWMMFL_SOCKET in the environment.

The problem is, FvwmCommand in da4387bd84c6dcf7e4483d545c48610e5420992b doesn't take FVWMMFL_SOCKET into account, resorting to guessing the socket path from FVWMMFL_SOCKET_PATH and DISPLAY. On my system it happens to guess *wrongly*: when the socket is created, it uses DISPLAY=:1 (as received from .Xsession), but under fvwm3, I have DISPLAY=:1.0 (with screen number). 

## Upfront Information

* Fvwm3 version (run: `fvwm3 --version`): fvwm3 1.1.1 (1.1.0-64-gf0366381)
with support for:  ReadLine, XPM, PNG, SVG, Shape, XShm, SM, Bidi text, XRandR, XRender, XCursor, XFT, NLS

* Linux distribution or BSD name/version: Ubuntu jammy

* Platform (run: `uname -sp`): Linux x86_64

## Expected Behaviour

Launched XTerm from FVWM, started FvwmCommand Module FvwmConsole, expected to see the console

## Actual Behaviour

```
Cannot find FvwmMFL socket /run/user/1000/fvwm/fvwm_mfl_:1.0.sock.
Please start FvwmMFL module from your fvwm configuration.
```

## Steps to Reproduce

How can the problem be reproduced?  
* Have screenless DISPLAY=:n in your fvwm3's starting environment
* Get DISPLAY=:n.0 in an XTerm launched by fvwm
* See how FvwmCommand fails to guess the socket path

akovalenko:
Follow-up to da4387bd FvwmMFL: introduce FVWMML_SOCKET_PATH for namespacing. It breaks FvwmCommand for me because instead of taking FVWMMFL_SOCKET from the environment, FvwmCommand guesses it from FVWMMFL_SOCKET_PATH (incorrectly, DISPLAY=:1.0 while socket being :1.sock)

When running FvwmCommand from a specific fvwm instance, FVWMMFL_SOCKET is available, so let FvwmCommand use it. In the "outer" environment where only FVWMML_SOCKET_PATH is available, let it resort to guessing socket by $DISPLAY. Fixes #1034.

### DIFF
--- a/bin/FvwmCommand.in
+++ b/bin/FvwmCommand.in
@@ -31,7 +31,9 @@ def mflclnt(verbose, read_from_stdin, info, fvwm_socket, args):
     #  + Use TMPDIR/fvwmmfl/fvwm_mfl_DISPLAY.sock if defined.
     #  + Last use default location /tmp/fvwmmfl/fvwm_mfl_DISPLAY.sock
     if not fvwm_socket:
-        if 'FVWMMFL_SOCKET_PATH' in os.environ:
+        if 'FVWMMFL_SOCKET' in os.environ:
+            fvwm_socket = str(os.environ.get('FVWMMFL_SOCKET'))
+        elif 'FVWMMFL_SOCKET_PATH' in os.environ:
             fvwm_socket = str(os.environ.get('FVWMMFL_SOCKET_PATH'))
             fvwm_socket += sock_file_name
         elif 'TMPDIR' in os.environ:
@@ -162,4 +164,3 @@ def main():
 
 if __name__ == "__main__":
     main()
-


ThomasAdam:
Damn... I missed that you'd done this, @akovalenko -- sorry.

Is this PR still required?

ThomasAdam:
Fixed via PR f6dd682f8c40c10803de81a6c65d91a37e764eff

PR #1046, opened 2024-07-09
delete -D cli arg

ileGITimo:
removes -D from --help
(Emacs added a \n at the eof)


somiaj:
@ileGITimo Note, if you add `editorconfig` support of emacs, it should honor the `.editorconfig` file and not add that extra new line.

PR #1047, opened 2024-07-10
Fix Restart functionality (after removal of single screen support)

cmeerw:
Removal of single screen support (#977) broke Restart functionality, as session.c still appended "-s screen_num" to the restart command args, resulting in the following error:

```
usage: fvwm3 [-d display] [-f cfgfile] [-c cmd] [-I vis-id | -C vis-class] [-l colors [-L|A|S|P] ...] [-r] [OTHER OPTIONS] ...
Try 'fvwm3 --help' for more information.
invalid option -- -s
```


ThomasAdam:
Thanks.

PR #1052, opened 2024-07-22
bring back negative desks, if ALLOW_NEGATIVE_DESKS is defined

ileGITimo:
- use INT_MIN instead of -1 to flag "special desks"
- check for "special desks" against INT_MIN for instead of == -1 or < 0
- two loops from 0 to ndesks use i+desk1 instead of i (no-op if desk1 is 0)
- new cpp defintion ALLOW_NEGATIVE_DESKS
- FvwmPager defines min_desk as 0, if above not defined, as INT_MIN+1 if defined
- defining ALLOW_NEGATIVE_DESKS will allow desks from INT_MIN+1 through INT_MAX, breaking EWMH
- not defining, the default, should not change any behavior

### DIFF
--- a/fvwm/add_window.c
+++ b/fvwm/add_window.c
@@ -2435,7 +2435,7 @@ FvwmWindow *AddWindow(
 	setup_key_and_button_grabs(fw);
 
 	/****** inform modules of new window ******/
-	fw->UpdateDesk = -1; /* Initialize to ensure desk update. */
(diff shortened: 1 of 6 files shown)


ThomasAdam:
Hi @ileGITimo,

Thanks for this.

I'm definitely not having this as a compile-time option, even if this is something I thought we should support.

I'm also not entirely sure if I want to support negative desks.

I think I'll close this for now.

PR #1053, opened 2024-07-22
Don't check final_pos against MAX/MIN_X_WINDOW_POSITION

ileGITimo:
- don't check final_pos against MAX/MIN_X_WINDOW_POSITION
- allows creating windows on virtual positions outside of those limits
- seems to run fine, or my use of fvwm3 doesn't ever hit a problem with this

### DIFF
--- a/fvwm/move_resize.c
+++ b/fvwm/move_resize.c
@@ -664,14 +664,6 @@ static int GetOnePositionArgument(
 		fvwm_debug(
 			__func__, "new position is out of range: %d",
 			final_pos);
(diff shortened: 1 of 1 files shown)


ThomasAdam:
Hi @ileGITimo 

What is the PR fixing, exactly?  If not, why is this useful?

ileGITimo:
> What is the PR fixing, exactly? If not, why is this useful?

The question I have why is this code there to begin with?  It breaks the code for very wide desktops and placing windows outside of (virtual) +/- 327867, which is possible and works fine if the code is not there.  However there may be other paths that need it.

PR #1055, opened 2024-07-24
adds PrintInfo Colorsets to print currenly defined Colorsets

ileGITimo:
Add Print Colorsets option.
Function print_colorsets just walks the Colorsets array and calls DumpColorsets on each entry.
Added to lib/Colorset.c, could also be added to fvwm/colorset.c. 
To do so one would need to add an extern int nColorsets to libs/Colorset.h, which seems to be missing, though it may not be needed.  In libs/Colorset.h the declaration of nColorsets is one of the two lines following a /* globals */ comment.

### DIFF
--- a/doc/fvwm3_manpage_source.adoc
+++ b/doc/fvwm3_manpage_source.adoc
@@ -3471,6 +3471,9 @@ If _verbose_ is one or greater the palette used by fvwm is printed. If
 you have a limited color palette, and you run out of colors, this
 command might be helpful.
 +
+_Colorsets_ which prints information about the colorsets currently in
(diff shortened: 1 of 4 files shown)


ThomasAdam:
Hi @ileGITimo 

Thanks -- but, I don't think this is worth.

ileGITimo:
Colorsets still don't work for me, or I still don't understand how they work.  This helps me check which colorsets are actually defined at the time.  I'm actually thinking on hooking this to creating a window that will show the colors, instead of printing the numbers.  It's fine if not in the released code, but I'll keep it on mine.

PR #1057, opened 2024-08-20
FvwmMFL: fix singletion/locking mechanism

### LINKED ISSUE #1054: FvmwMFL did not get started because of wrong PID in pid file
Thanks for reporting your bug here!  The following template will help with
giving as much information as possible so that it's easier to diagnose and
fix.

## Upfront Information

the output.

```
[~]$ fvwm3 --version
fvwm3 1.0.6a (released)
with support for:  ReadLine, XPM, PNG, SVG, Shape, XShm, SM, Bidi text, XRandR, XRender, XCursor, XFT, NLS

fvwm3 comes with NO WARRANTY, to the extent permitted by law. You may
redistribute copies of fvwm under the terms of the GNU General Public License.
For more information about these matters, see the file named COPYING.
```

Debian testing

* Platform (run: `uname -sp`)

```
[~]$ uname -sp
Linux unknown
```

## Expected Behaviour

Module FvwmMFL gets started during FVWM startup.  My startup function:

```
AddToFunc StartFunction
+ I Echo StartFunction
+ I Module FvwmEvent
+ I Module FvwmMFL
```

## Actual Behaviour

Module FvwmMFL did not start.  This one time, I have never seen that previously.  And I could not reproduce it, so I can only provide post-mortem information.

In my journal I see these messages here (note that FVWM runs in debug mode and logs to the journal):

```
Jul 23 10:24:48 host01 systemd[1869]: Started fvwm3.scope - /usr/bin/fvwm3 -v -o -.
Jul 23 10:24:48 host01 fvwm3[1889]: [1721723088.820074] FScreenInit: Using RandR 1.6
Jul 23 10:24:48 host01 fvwm3[1889]: [1721723088.946628] main: Loading window states via (null)
Jul 23 10:24:48 host01 fvwm3[1889]: [1721723088.961279] parse_and_set_window_style: Bad style option: Color lightgrey/dimgrey
Jul 23 10:24:48 host01 fvwm3[1889]: [1721723088.961301] parse_and_set_window_style: Bad style option: HilightFore black
Jul 23 10:24:48 host01 fvwm3[1889]: [1721723088.961303] parse_and_set_window_style: Bad style option:  HilightBack grey
Jul 23 10:24:48 host01 fvwm3[1889]: loaded [0]: /home/farblos/.fvwm/config
Jul 23 10:24:48 host01 fvwm3[1889]: [1721723088.973442] CMD_Echo: StartFunction
Jul 23 10:24:48 host01 fvwm3[1971]: FvwmMFL is already running
```

So the module thinks it is already running.  Its pid file at that time looked like this:

```
[~]$ find . -iname '*fvwm*mfl*'
./.fvwm/fvwm_mfl_:0.pid
[~]$ ls -al ./.fvwm/fvwm_mfl_:0.pid
-rw-r--r-- 1 farblos farblos 4 May 21 10:20 ./.fvwm/fvwm_mfl_:0.pid
[~]$ cat ./.fvwm/fvwm_mfl_:0.pid
1889
```

What I do not understand:

- That file has a timestamp from May 21.  I have been using FVWM almost daily since then.
- It refers to PID 1889, which (see above journal) is the PID of the main FVWM process, and not the one of FvwmMFL.

I also tried restarting FVWM (from within FVWM), but that didn't help since the newly started FVWM process obviously inherits the PID of the old one:

```
Jul 23 11:03:23 host01 fvwm3[1889]: [1721725403.952478] FScreenInit: Using RandR 1.6
Jul 23 11:03:24 host01 fvwm3[1889]: [1721725404.097550] main: Loading window states via /home/farblos/.fvwm/.fs-restart-host01:0
Jul 23 11:03:24 host01 fvwm3[1889]: [1721725404.118143] parse_and_set_window_style: Bad style option: Color lightgrey/dimgrey
Jul 23 11:03:24 host01 fvwm3[1889]: [1721725404.118168] parse_and_set_window_style: Bad style option: HilightFore black
Jul 23 11:03:24 host01 fvwm3[1889]: [1721725404.118171] parse_and_set_window_style: Bad style option:  HilightBack grey
Jul 23 11:03:24 host01 fvwm3[1889]: loaded [0]: /home/farblos/.fvwm/config
Jul 23 11:03:24 host01 fvwm3[1889]: [1721725404.139117] CMD_Echo: StartFunction
Jul 23 11:03:24 host01 fvwm3[5340]: FvwmMFL is already running
```

Only after manually removing the PID file and then restarting FVWM would the FvwmMFL module start.

Some funny race condition during creation of the PID file?  No further ideas here.  If you don't have any ideas, either, feel free to close this issue as "not reproducible" ...

Thanks!

ThomasAdam:
There can only ever be one single instance of FvwmMFL running per FVWM instance/$DISPLAY.

Currently, this is done with logging the PID of the running FvwmMFL process and using that to look up a process with the same PID.  This is fragile though as the pid in the saved file might not be the samae PID of FvwmMFL -- therefore, FvwmMFL wouldn't start.

Instead, switch to using a lock file which is predicated per $DISPLAY -- achieving the same result, but less error prone.

Fixes #1054

### DIFF
--- a/modules/FvwmMFL/FvwmMFL.c
+++ b/modules/FvwmMFL/FvwmMFL.c
@@ -19,6 +19,7 @@
 #include "libs/envvar.h"
 #include "libs/cJSON.h"
 
+#include <sys/file.h>
 #include <sys/socket.h>
 #include <sys/un.h>
 #include <sys/time.h>
@@ -52,7 +53,8 @@
 static int debug;
 struct fvwm_msg;
 static char *socket_name, *socket_basepath;
-static char *pid_file;
+static char *lockfile = NULL;
+static int fvwmmfl_lock_fd = -1;
 
 struct client {
 	struct bufferevent	*comms;
@@ -122,11 +124,8 @@ static struct fvwm_msg *fvwm_msg_new(void);
 static void fvwm_msg_free(struct fvwm_msg *);
 static void register_interest(void);
 static void send_version_info(struct client *);
-static void set_socket_pathname(void);
-static int check_pid(void);
-static void set_pid_file(void);
-static void create_pid_file(void);
-static void delete_pid_file(void);
+static void set_pathnames(void);
+static int lock_fvwmmfl(void);
 static void fm_to_json(struct fvwm_msg *);
 
 static void
@@ -137,88 +136,6 @@ fm_to_json(struct fvwm_msg *fm)
 		cJSON_Delete(fm->j_obj);
 	}
 }
-
-static void
-delete_pid_file(void)
-{
-	if (pid_file == NULL)
-		return;
-
-	unlink(pid_file);
-}
-
-static void
-set_pid_file(void)
-{
-	char	*dsp = getenv("DISPLAY");
-
-	if (dsp == NULL) {
-		fprintf(stderr,
-		    "DISPLAY is not set in the environment\n");
-		exit(1);
-	}
-
-	free(pid_file);
-	xasprintf(&pid_file, "%s/fvwm_mfl_%s.pid", socket_basepath, dsp);
-}
-
-static void
-create_pid_file(void)
-{
-	FILE	*pf;
-
-	if ((pf = fopen(pid_file, "w")) == NULL) {
-		fprintf(stderr, "Couldn't open %s because: %s\n", pid_file,
-			strerror(errno));
-		exit(1);
-	}
-	fprintf(pf, "%d", getpid());
-
-	if (fclose(pf) != 0) {
-		fprintf(stderr, "Couldn't close %s because: %s\n", pid_file,
-			strerror(errno));
-		exit(1);
-	}
-
-	/* Alongside sigTerm, ensure we remove the pid when exiting as well. */
-	atexit(delete_pid_file);
-}
-
-static int
-check_pid(void)
-{
-	FILE	*pf;
-	int	 pid;
-
-	if (pid_file == NULL)
-		set_pid_file();
-
-	if (access(pid_file, F_OK) != 0) {
-		create_pid_file();
-		return (1);
-	}
-
-	if ((pf = fopen(pid_file, "r")) == NULL) {
-		fprintf(stderr, "Couldn't open %s for reading: %s\n", pid_file,
-			strerror(errno));
-		exit(1);
-	}
-	if (fscanf(pf, "%d", &pid) != 1) {
-		fprintf(stderr, "Error reading pid from %s: %s\n", pid_file,
-			strerror(errno));
-		exit(1);
-	};
-
-	/* Non-fatal if we can't close this file handle from reading. */
-	(void)fclose(pf);
-
-	if ((kill(pid, 0) == 0)) {
-		fprintf(stderr, "FvwmMFL is already running\n");
-		return (0);
-	}
-	return (1);
-}
-
 
 static struct fvwm_msg *
 fvwm_msg_new(void)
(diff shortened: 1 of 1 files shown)


modules/FvwmMFL/FvwmMFL.c
```diff
@@ -246,7 +163,7 @@ static void
 HandleTerminate(int fd, short what, void *arg)
 {
 	fprintf(stderr, "%s: dying...\n", __func__);
-	delete_pid_file();
```
farblos, line 249: Probably close `fvwmmfl_lock_fd` here? And if only to show that you have not forgotten :-).  Also in the other place(s) where you unlink.

modules/FvwmMFL/FvwmMFL.c
```diff
@@ -849,6 +766,31 @@ set_socket_pathname(void)
 
 	xasprintf(&socket_name, "%s/fvwm_mfl_%s.sock", socket_basepath,
 		display_env);
+
```
farblos, line 769: What about naming the file `fvwm_mfl_%s.lock` instead?  But that's purely cosmetical...

somiaj: This also puts all the sockets (if fvwm is running on multiple DISPLAY's at the same time) in a single directory, so a bit for organization too. But as you said mostly cosmetical, I'm just stating the logic of this choice.

modules/FvwmMFL/FvwmMFL.c
```diff
@@ -246,7 +163,7 @@ static void
 HandleTerminate(int fd, short what, void *arg)
 {
 	fprintf(stderr, "%s: dying...\n", __func__);
-	delete_pid_file();
+	unlink(lockfile);
 	unlink(socket_name);
 	rmdir(socket_basepath);
```
farblos, line 168: Oh, and you probably should not just replace `delete_pid_file()` by `unlink(lockfile)`, but also complement all calls to `unlink(socket_name)` by a call to `unlink(lockfile)` (plus closing the lok file handle)? Or did I miss something here?

modules/FvwmMFL/FvwmMFL.c
farblos: With Thomas' proposal, you would have the files in `ls` output grouped by type, with my proposal grouped by display.  Not sure what is more meaningful.

farblos: Oh, to clarify: I mean using `"%s/fvwm_mfl_%s.lock"` (with the leading `%s/` of course) instead of using `"%s/fvwmmfl_lockfile_%s.lock"`.

somiaj: Thanks for the clarification, though this is the location of the socket file, so I would use `.sock` and not `.lock`.

modules/FvwmMFL/FvwmMFL.c
```diff
@@ -849,6 +766,31 @@ set_socket_pathname(void)
...
+lock_fvwmmfl(void)
+{
+	struct flock	l;
+	int		fd = -1;
+
+	if ((fd = open(lockfile, O_CREAT|O_WRONLY, 0600)) == -1)
+		return fd;
+	l.l_start = 0;
+	l.l_type = F_WRLCK;
+	l.l_len = 0;
```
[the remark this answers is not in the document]
ThomasAdam: Yes, it's not a problem.

modules/FvwmMFL/FvwmMFL.c
ThomasAdam: It's not changing.

PR #1062, opened 2024-10-03
Locale translation file updates.

somiaj:
Move the scripts to update and build the LC_MESSAGE localization files out of automake makefiles (these are not used in the build anyways, but must be manually ran by translators). The automake file now only installs/uninstalls the LC_MESSAGE files for each language.

There is a new script, po/fvwmpo.sh, which can be used to update the main pot file, initialize, update, and build the language files for the strings that gettext translates. This does essentially the same thing the make targets use to.

Move all the translations into a single fvwm3.pot file (which merges the FvwmScript.pot file into the fvwm3.pot file). The file has also been renamed from fvwm.pot to fvwm3.pot to allow it to be installed along size fvwm2.

Last, this updates the template, and merges all translation files to be in line with the current template, and rebuilds all the .gmo files that get installed into the locale directory.


po/fvwmpo.sh
```diff
@@ -0,0 +1,97 @@
+#!/bin/sh
+FVWMPOT=fvwm3.pot
+MSGINIT=/usr/bin/msginit
+MSGFMT=/usr/bin/msgfmt
+MSGUNIQ=/usr/bin/msguniq
+MSGMERGE=/usr/bin/msgmerge
+XGETTEXT=/usr/bin/xgettext
+
```
ThomasAdam, line 8: I know this is lifted from the Makefile, but I think we should drop the path-prefixes to the binaries, and let $PATH handle these.

Especially when you consider gettext on *BSD is likely to use /usr/local/bin as the location.

ThomasAdam (changes requested):
Hi @somiaj 

All looks good, save for one small change to the po/fvwmpo.sh script.

po/fvwmpo.sh
somiaj: Full paths have been removed.

somiaj:
I also found that the build command for building the `.gmo` files was excluding fuzzy translations, so I added the `-f` flag to this, so fuzzy translations are now included. Now in my testing all strings in the `.po` files seems to be correctly translated.

PR #1067, opened 2024-10-13
Adjust windows to fit working area after move.

### LINKED ISSUE #1033: FvwmRearrange broken on multi-monitor layout
## Upfront Information

When using a 3 monitor configuration with following layout:

```
Screen 0: minimum 8 x 8, current 3840 x 1980, maximum 32767 x 32767
HDMI-0 connected primary 1920x1080+0+900 (normal left inverted right x axis y axis) 700mm x 400mm
DP-2 connected 1920x1080+1920+900 (normal left inverted right x axis y axis) 700mm x 400mm
DP-5 connected 1600x900+1117+0 (normal left inverted right x axis y axis) 443mm x 249mm
```
Invoking

```
 `FvwmCommand FvwmRearrange -tile
````````
`
`
Causes windows to go off-screen somewhere below the top monitor rather than being properly tiled.
Similar issues happen with the cascade option.

fvwm3 1.1.1 (dfc46ddc) (latest main branch at the time of writing this issue)

Gentoo GNU+Linux

* Platform (run: `uname -sp`)

Linux AMD Ryzen 9 3950X 16-Core Processor

## Expected Behaviour

1) I expected windows to be tiled within a visible area of the monitor.
2) This may not be the actual behavior of FvwmRearrange -tile, but I was expecting (and hoping) to tile only the windows on the active screen/monitor (the one the mouse cursor is on) and not the all windows globally, and I expected them to be tiled ON that active monitor/screen.

## Actual Behaviour

Chaos ensued. ;)
Windows vanished, ending up somewhere below the top-most monitor, yet somehow also not visible on either monitor below it. I could just barely see the top right corner of them in the bottom left corner of the top screen.

## Enabling logging

Invoking FvwmRearrange is not logging anything additional to the debug log, but here's the randr debug bits.
```

[1718059140.075366] monitor_dump_state: Monitor Debug
[1718059140.075375] monitor_dump_state: 	number of outputs: 3
[1718059140.075382] monitor_dump_state: 	Number:	0
	Name:	DP-5
	Disabled:	false
	Is Primary:	no
	Is Current:	no
	Is Previous:	no
	Output:	559
	Coords:	{x: 1117, y: 0, w: 1600, h: 900}
	VirtScr: {
		VxMax: 0, VyMax: 0, Vx: 0, Vy: 0
		EdgeScrollX: 0, EdgeScrollY: 0
		CurrentDesk: 0
		CurrentPage: {x: 0, y: 0}
		MyDisplayWidth: 3840, MyDisplayHeight: 1980
	}
	EWMH: {
		BaseStrut Top:    0
		BaseStrut Bottom: 0
		BaseStrut Left:   0
		BaseStrut Right:  0
	}
	Desktops:	yes
	Flags:global

[1718059140.075394] monitor_dump_state: 	Number:	1
	Name:	HDMI-0
	Disabled:	false
	Is Primary:	yes
	Is Current:	no
	Is Previous:	no
	Output:	446
	Coords:	{x: 0, y: 900, w: 1920, h: 1080}
	VirtScr: {
		VxMax: 0, VyMax: 0, Vx: 0, Vy: 0
		EdgeScrollX: 0, EdgeScrollY: 0
		CurrentDesk: 0
		CurrentPage: {x: 0, y: 0}
		MyDisplayWidth: 3840, MyDisplayHeight: 1980
	}
	EWMH: {
		BaseStrut Top:    0
		BaseStrut Bottom: 0
		BaseStrut Left:   0
		BaseStrut Right:  0
	}
	Desktops:	yes
	Flags:global

[1718059140.075416] monitor_dump_state: 	Number:	2
	Name:	DP-2
	Disabled:	false
	Is Primary:	no
	Is Current:	yes
	Is Previous:	no
	Output:	554
	Coords:	{x: 1920, y: 900, w: 1920, h: 1080}
	VirtScr: {
		VxMax: 0, VyMax: 0, Vx: 0, Vy: 0
		EdgeScrollX: 0, EdgeScrollY: 0
		CurrentDesk: 0
		CurrentPage: {x: 0, y: 0}
		MyDisplayWidth: 3840, MyDisplayHeight: 1980
	}
	EWMH: {
		BaseStrut Top:    0
		BaseStrut Bottom: 0
		BaseStrut Left:   0
		BaseStrut Right:  0
	}
	Desktops:	yes
	Flags:global

[1718059140.081925] ewmh_ComputeAndSetWorkArea: monitor 'HDMI-0': {l: 0, r: 120, t: 0, b: 0} {x: 0, y: 0, w: 1800, h: 1080}
[1718059140.081975] ewmh_ComputeAndSetWorkArea: monitor 'DP-2': {l: 0, r: 120, t: 0, b: 0} {x: 0, y: 0, w: 1800, h: 1080}
[1718059140.082016] ewmh_ComputeAndSetWorkArea: monitor 'DP-5': {l: 0, r: 120, t: 0, b: 0} {x: 0, y: 0, w: 1480, h: 900}
[1718059140.302562] parse_and_set_window_style: Bad style option:  NoBorders
[1718059140.302620] parse_and_set_window_style: Bad style option:  NoBorders
[1718059140.302663] parse_and_set_window_style: Bad style option:  NoBorders
[1718059140.307615] HandleMapRequestKeepRaised: Window 0x440000a is new and is being recaptured and in an iconic state. Sending fake map/unmap
[1718059140.347200] ewmh_ComputeAndSetWorkArea: monitor 'DP-5': {l: 0, r: 120, t: 18, b: 0} {x: 0, y: 18, w: 1480, h: 882}
[1718059145.667876] monitor_dump_state: Monitor Debug
[1718059145.667897] monitor_dump_state: 	number of outputs: 3
[1718059145.667901] monitor_dump_state: 	Number:	2
	Name:	DP-2
	Disabled:	false
	Is Primary:	no
	Is Current:	yes
	Is Previous:	no
	Output:	554
	Coords:	{x: 1920, y: 900, w: 1920, h: 1080}
	VirtScr: {
		VxMax: 3840, VyMax: 1980, Vx: 3840, Vy: 0
		EdgeScrollX: 0, EdgeScrollY: 0
		CurrentDesk: 0
		CurrentPage: {x: 1, y: 0}
		MyDisplayWidth: 3840, MyDisplayHeight: 1980
	}
	EWMH: {
		BaseStrut Top:    0
		BaseStrut Bottom: 0
		BaseStrut Left:   0
		BaseStrut Right:  120
	}
	Desktops:	yes
	Flags:per-monitor

```

somiaj:
When moving windows, it made sense to consider the working area base struts before the move when the base struts were on the global view port. Now that the base struts are per monitor this no longer makes sense, and caused strange behavior when computing where to move a window.

Instead the window's position should be adjusted to fit inside the working area after the move is done. This way the position of the window is always computed relative to the global screen (unless the 'screen RANDR_NAME' option is provided).

Notes, this is a breaking change as many multiple monitor configs have updated to work around this behavior. This should also fix #1033 and other issues where computations are done relative to the global screen but incorrectly adjusted when ran on monitors not located at `(0,0)`.

This should also be tested, I've only done limited testing so far, but all my tests have been favorable. Hopefully anything this breaks is due to having to work around the old incorrect behavior.

### DIFF
--- a/doc/fvwm3_manpage_source.adoc
+++ b/doc/fvwm3_manpage_source.adoc
@@ -3906,10 +3906,16 @@ Move [screen _S_] [desk _N_] [w | m | v]_x_[p | w] [w | m | v]_y_[p | w] [Warp] 
 ....
 +
 This will move the window to the _x_ and _y_ position (see below).
-By default, the EWMH working area is honoured. If he trailing option
-_ewmhiwa_ is given, then the window position will ignore the working
-area (such as ignoring any values set via *EwmhBaseStruts*). If the
-option _Warp_ is given then the pointer is warped to the window.
+By default, the EWMH working area of each monitor is honoured (the
+working area for each monitor is set via *EwmhBaseStruts* and honors
+any strut hints provided by windows on the monitor). This means that
+if a window is placed outside the working area, the position of the
+window will be adjusted to fit inside the working area of the screen
+the center of the window is on. If the trailing option _ewmhiwa_ is
+given, then the window position will ignore the working area and its
+position will not be adjusted (this option is needed to move windows
+off the screen and to have full control of where it is placed). If
+the option _Warp_ is given then the pointer is warped to the window.
 +
 If the literal option _screen_ followed by a RandR screen name _S_ is
 specified, the coordinates are interpreted as relative to the given
@@ -3925,15 +3931,17 @@ is updated to be the same as the new monitor. This option can override
 that behavior by specifying which desk the window should end up on.
 +
 The positional arguments _x_ and _y_ can specify an absolute or relative
-position from either the left/top or right/bottom of the screen. By default,
-the numeric value given is interpreted as a percentage of the screen
-width/height, but a trailing '_p_' changes the interpretation to mean pixels,
-while a trailing '_w_' means percent of the window width/height. To move the
-window relative to its current position, add the '_w_' (for "window") prefix
-before the _x_ and/or _y_ value. To move the window to a position relative to
-the current location of the pointer, add the '_m_' (for "mouse") prefix. To
-move the window relative to the virtual screen coordinates, add the '_v_'
-(for "virtual screen") prefix. This is mostly for internal use with FvwmPager,
+position from either the left/top (positive values) or right/bottom
+(negative values) of the global screen (the bounding box that contains all
+monitors) or specified _screen_. By default, the numeric value given is
+interpreted as a percentage of the screen's width/height, but a trailing
+'_p_' changes the interpretation to mean pixels, while a trailing '_w_'
+means percent of the window width/height. To move the window relative to
+its current position, add the '_w_' (for "window") prefix before the _x_
+and/or _y_ value. To move the window to a position relative to the current
+location of the pointer, add the '_m_' (for "mouse") prefix. To move the
+window relative to the virtual screen coordinates, add the '_v_' (for
+"virtual screen") prefix. This is mostly for internal use with FvwmPager,
 but can be used to give exact coordinates on the virtual screen and is best
 used with the '_p_' suffix. To leave either coordinate unchanged, "_keep_"
 can be specified in place of _x_ or _y_.
--- a/fvwm/ewmh.c
+++ b/fvwm/ewmh.c
@@ -1116,9 +1116,9 @@ void EWMH_UpdateWorkArea(struct monitor *m)
 }
 
 void EWMH_GetWorkAreaIntersection(
-	FvwmWindow *fw, int *x, int *y, int *w, int *h, int type)
+	struct monitor *mon, int *x, int *y, int *w, int *h, int type)
 {
-	struct monitor	*m = (fw && fw->m) ? fw->m : monitor_get_current();
+	struct monitor	*m = (mon) ? mon : monitor_get_current();
 
 	EWMH_UpdateWorkArea(m);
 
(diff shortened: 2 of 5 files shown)


somiaj:
This is another example of issues that should be fixed by this PR.

https://fvwmforums.org/t/fvwm3-setting-up-one-desktop-for-three-monitors/4734/5

somiaj:
When working with the previous patch, I noticed that the `Move shuffle` commands didn't take into consideration any window hints when computing the working area. Added a patch to fix that issue as well.

somiaj:
A few more notes, move coordinates are always done relative to the global screen (or specified screen) and are never relative to the working area (as it was before, so that is the breaking change). So users who want to move relative to the working area have to take that into account with their move commands. Instead the window's position is adjusted to ensure it is inside the working area.

Second, gravity isn't taken into account (probably it should be), and it is always the top left corner that determines what screen the window is on, and if a window is bigger than the working area, the position is adjusted so the top/left corner is at the edge of the working area.

somiaj:
Updated so the center of the window (not the top/left corner) is used to determine which monitor the window is on. This way if a move command moves the top/left corner slightly off a monitor's boundary, it will still be adjusted to be on the monitor with the most overlap.

jnse:
I can confirm that using this branch, FvwmRearrange no longer moves windows off screen. There is still a few oddities, probably more to do with the module itself than this change:

* 'FvwmRearrange -tile -h' seems to be adding unexpected padding between the windows.
* 'FvwmRearrange -tile -desk' moves all windows to the current desk rather than tiling the windows on the desk they were originally on (which is what I'd expect it to do, although I'm not sure if it behaved like that in the past)

somiaj:
@jnse thanks, those issues are going to be addressed in a future pull request, as this is only updating how fvwm Move command works internally (which FvwmRearrange) uses. I am working on improving the logic of FvwmRearrange to know things what the base struts are, which desk you are on, which page, and so on.

@ThomasAdam I think this is probably ready to go, as it fixed the primary issue with modules moving windows relative to base struts in multimonitor situation.

PR #1069, opened 2024-10-14
Improve monitor event handling

### LINKED ISSUE #266: FvwmEvent monitor_* events are not processing their functions correctly.
By using FvwmEvent to Echo some string into fvwm log, I get "-1" instead of that string from the function.
As a base FvwmEvent configuration, NsCDE fvwm3_support branch is used. On this config:

https://github.com/NsCDE/NsCDE/blob/fvwm3_support/NsCDE/config/NsCDE-Event.conf

new config is appended for testing:

```
DestroyFunc f_EventTestME
AddToFunc f_EventTestME
+ I Echo EVENT DEBUG ME ============================> $*

DestroyFunc f_EventTestMD
AddToFunc f_EventTestMD
+ I Echo EVENT DEBUG MD ============================> $*

DestroyFunc f_EventTestMC
AddToFunc f_EventTestMC
+ I Echo EVENT DEBUG MC ============================> $*

DestroyFunc f_EventTestMF
AddToFunc f_EventTestMF
+ I Echo EVENT DEBUG MF ============================> $*

*MainLoop: monitor_enabled f_EventTestME __me__
*MainLoop: monitor_disabled f_EventTestMD __md__
*MainLoop: monitor_changed f_EventTestMC __mc__
*MainLoop: monitor_focus f_EventTestMF __mf__

```

This is the result:

[1603230503.629785] CMD_Echo: EVENT DEBUG MD ============================> -1
[1603230544.602999] CMD_Echo: EVENT DEBUG ME ============================> -1
[1603230250.458667] CMD_Echo: EVENT DEBUG MF ============================> -1

Additionally, it looks like monitor_focus is more often not triggered when it has to be then triggered. Moving pointer between the screens does not produce the output in log, while moving it to FvwmButtons based FrontPanel which has different Style options from other windows succeeds almost every time to produce output in log.

farblos:
### **What does this PR do?**

After the monitor event handling changes earlier this year I noticed that I do not get any longer `monitor_changed` events.  I started checking the relevant code and found some issues, then some more, and then I went down the rabbit hole ...

Know I have Fvwm in a state where I get `montor_changed`, `_disabled`, `_enabled` events exactly mirroring the xrandr changes.

Here is a rough overview of what I did:

- I removed old code related to `monitorsold_q`, since I decided that it is a left-over of an earlier event implementation
- I fixed some conditions on `m->emit` in `monitor_emit_broadcast`, plus I removed a `break` statement that would (IMO) exit the loop even if there are still events that need reporting (also some left-over?)
- I added some check in `dispatch_event` to avoid double handling of identical events (even though `scan_monitors` should now handle these well)
- I extended the event handling in `scan_monitors` to more accurately detect changes in monitor configuration also in all sorts of edge cases

### **Known Open Issues and Questions**

- The handling of events on primary monitor changes is still not quite correct. It probably should be also redone in terms of some flags on `m->emit`.  Not sure how relevant that is, though.
- Any idea why I get duplicate RandR events with identical serial number? Anybody else seing this?

### Edits

- Removed the draft header.
- Updated overview and known issues.
- Removed the link to that other issue.

### DIFF
--- a/libs/FScreen.c
+++ b/libs/FScreen.c
@@ -65,7 +65,6 @@ enum monitor_tracking monitor_mode;
 bool			 is_tracking_shared;
 struct screen_infos	 screen_info_q, screen_info_q_temp;
 struct monitors		 monitor_q;
-struct monitorsold	 monitorsold_q;
 int randr_event;
 const char *prev_focused_monitor;
 static struct monitor	*monitor_global = NULL;
@@ -348,17 +347,6 @@ monitor_by_last_primary(void)
 	return (m);
 }
 
-int changed_monitor_count(void)
-{
-	struct monitor	*m;
-	int c = 0;
-
-	TAILQ_FOREACH(m, &monitorsold_q, oentry)
-		c++;
-
-	return (c);
-}
-
 static void
 monitor_check_primary(void)
 {
@@ -510,7 +498,6 @@ monitor_mark_changed(struct monitor *m, XRRMonitorInfo *rrm)
 	    m->si->w != rrm->width ||
 	    m->si->h != rrm->height) {
 		m->flags |= MONITOR_CHANGED;
-		m->emit |= MONITOR_CHANGED;
 
 		fvwm_debug(__func__, "%s: x: %d, y: %d, w: %d, h: %d "
 			"comp: x: %d, y: %d, w: %d, h: %d\n",
@@ -518,7 +505,6 @@ monitor_mark_changed(struct monitor *m, XRRMonitorInfo *rrm)
 			rrm->x, rrm->y, rrm->width, rrm->height);
 
 		monitor_set_coords(m, *rrm);
-		TAILQ_INSERT_TAIL(&monitorsold_q, m, oentry);
 
 		return (true);
 	}
@@ -550,6 +536,10 @@ scan_screens(Display *dpy)
     	int			 i, n = 0;
 	Window			 root = RootWindow(dpy, DefaultScreen(dpy));
 
+	RB_FOREACH(m, monitors, &monitor_q) {
+		m->flags &= ~(MONITOR_NEW|MONITOR_CHANGED);
+	}
+
 	rrm = XRRGetMonitors(dpy, root, true, &n);
 	if (n <= 0 && (!randr_initialised && monitor_get_count() == 0)) {
 		fvwm_debug(__func__, "get monitors failed\n");
@@ -578,9 +568,12 @@ scan_screens(Display *dpy)
 	 *   - It's an existing monitor (position changed)
 	 *   - It's an existing monitor which has been toggled on or off.
 	 *
-	 *   In such cases, we must detect if the monitor exists and what
-	 *   state it is in.
+	 * In such cases, we must detect if the monitor exists and what state it
+	 * is in.  Regardless of how the monitor state has changed, we flag all
+	 * monitors reported by XRRGetMonitors with flag MONITOR_FOUND, which we
+	 * use below to determine its new state.
 	 */
+        fvwm_debug(__func__, "Case 2: processing %d monitors", n);
 	for (i = 0; i < n; i++) {
 		if ((name = XGetAtomName(dpy, rrm[i].name)) == NULL)
 			continue;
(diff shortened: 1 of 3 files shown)


ThomasAdam:
Thanks, @farblos 

There's a few changes here I don't think we need, plus many changes we can consolidate. 

We don't need any of the documentation changes you've made. 

I'll reply more fully in the coming days. The needs a lot more refinement yet.

farblos:
>I'll reply more fully in the coming days. The needs a lot more refinement yet.

Sure, thanks. If you think it'll help we can discuss this also via IRC.  Only we need to arrange some kind of a meeting then, since I'm usually not available on IRC.

ThomasAdam:
Thanks, @farblos -- I'm always in #fvwm on IRC, so if you're not always around, catch me when you can... :)

libs/FScreen.c
```diff
@@ -577,50 +587,85 @@ scan_screens(Display *dpy)
...
+		int  flags = m->flags;
+		Bool found = m->flags & MONITOR_FOUND;
+
 
 		/* Check for monitor connection status -- whether a monitor is
 		 * active or not.  Clearing the MONITOR_FOUND flag is
 		 * important here so that the monitor is reconsidered again.
+		 *
+		 * Better try to keep these conditions readable than
+		 * ingenious ...
```
ThomasAdam, line 636: Remove.

libs/FScreen.c
```diff
@@ -577,50 +587,85 @@ scan_screens(Display *dpy)
...
+		/* This case happens if !randr_initialised. */
+		else if (!found && (flags & MONITOR_NEW)) {
+			m->emit   = 0;
+		}
+		else if (!found && (flags & MONITOR_DISABLED)) {
+			m->emit   = 0;
+		}
+		else /* (!found && (! (flags & MONITOR_DISABLED))) */ {
+			m->flags |= MONITOR_DISABLED;
+			m->emit   = MONITOR_DISABLED;
```
ThomasAdam, line 663: If you're suggesting this, please cuddle the else statements:

```
else if (..) {
...
} else if (..._ {
}
```

libs/FScreen.c
```diff
@@ -577,50 +587,85 @@ scan_screens(Display *dpy)
...
 		}
 		m->flags &= ~MONITOR_FOUND;
+		fvwm_debug(__func__, "Flags: %s -> %04x, %04x\n", m->si->name, m->flags, m->emit);
```
ThomasAdam, line 666: Remove.

libs/FScreen.c
```diff
@@ -577,50 +587,85 @@ scan_screens(Display *dpy)
...
 	}
+
```
ThomasAdam, line 668: Remove.

modules/FvwmEvent/FvwmEvent.c
```diff
@@ -63,6 +63,13 @@
 typedef struct
 {
 	char *name;
+	/* -1: no action arguments
+	 *  0: action argument is window ID
+	 *  {0, 1, 2} | ARG_NO_WINID:
+	 *     action argument is a signed long in the given data slot
+	 *  0 | ARG_MONITOR | ARG_NO_WINID:
+	 *     action argument is monitor info
+	 */
```
ThomasAdam, line 72: Remove.

modules/FvwmEvent/FvwmEvent.c
```diff
@@ -94,7 +101,8 @@ static Bool PassID = False;
 static Bool audio_compat = False;
 static char *audio_play_dir = NULL;
 
-#define ARG_NO_WINID 1024  /* just a large number */
+#define ARG_NO_WINID 2<<10  /* just a large number */
+#define ARG_MONITOR  2<<11  /* ditto */
```
ThomasAdam, line 105: Why are you changing this?  Also, no need for the bit-shifts here.

ThomasAdam (changes requested):
Hi @farblos 

Thanks for this.

Overall it looks good to me.  I've left specific comments in this review.

Key things here to note are:

1. I'd like you to remove the comments you've added before each function in libs/FScreen.c -- presumably it's there to help you, but doesn't belong in this commit.
2. We don't need MONITORS.md so that can go.
3. We need to think more carefully about FvwmEvent, and what we want to pass to it.  That said, I don't want us to introduce a `PassArgs` component to FvwmEvent as you're suggesting.  I suspect it will be enough to just send m->si->name to either RandRFunc, for the specific event in FvwmEvent.

If you could make the required changes, this should help reduce some of the extraneous noise in this change.

Any questions, please let me know.

farblos:
Since we settled on `RandRFunc` in IRC, I'll revert all changes on `FvwmEvent.c` and its man page.

Regardless of that I might (or not) take another stab at the man page in a separate PR. I think it could state more clearly and at the beginning how exactly the command is built. For example, I took me reading the sources to understand that I can leave `*FvwmEvent: Cmd` blank ...

farblos:
Hi @ThomasAdam,

I think I'm through with the changes from your review, removing most of the noise.  Plus I reverted all changes related to `FvwmEvent`, keeping basically only:

- removal of obsolete code related to `monitorsold_q`
- consolidation of event handling and emitting code

I removed the "draft" flag from the PR, but I'm not sure how to close the open review status ...

Anyway, we can continue from here as you prefer:

- I could re-request review
- I could squash and rebase the commits
- You could add your patches on top as you have been planning
- You could even merge the PR as is, as this a) seems to be a self-contained unit of work and b) comprises already what is most important to me

Just pls let me know how to continue, and preferredly in this PR, since I won't hang around in IRC the next days.

Thanks!

ThomasAdam:
@farblos 

Thanks.  Please have a look at commit 652eb3492175ab1984dc07a7f1dd43acba818338 on branch `ta/review-gh-1069`.

I've squashed all of your intermediary commits into just one commit for these changes (even though these changes cross fvwm <-> module boundaries, it's  all related).  I've left the authorship intact, but changed the commit message.

You'll note in I've left a commented block with `FIXME` in it -- this is in `monitor_update_ewmh()`.  Please take a look.

In order to see what I've changed, relative to this PR, you can do:

```
git diff improve-monitor-events..ta/review-gh-1069
```

Which assumes your branch for this PR is called `improve-monitor-events`.

What I suggest you do is cherry-pick 652eb3492175ab1984dc07a7f1dd43acba818338 onto your branch, make any changes to that commit, and just force-push it to this PR.

Hope that helps.  Any questions let me know.

farblos:
Fair enough, thanks.

Regarding your FIXME in `monitor_update_ewmh`.  The code in `main` at that point does the following:

```c
TAILQ_FOREACH(mo, &monitorsold_q, oentry) {
  if (strcmp(m->si->name, mo->si->name) != 0)
    continue;
  if (mo->Desktops == NULL)
    continue;
  for (t = Scr.FvwmRoot.next; t; t = t->next) {
    if (t->m == mo) {
      t->m = m;
      update_fvwm_monitor(t);
    }
  }
  m->emit |= MONITOR_CHANGED;
}
```

Since you have let me remove the `monitorsold_q`-related code, I guess it really must have been obsolete, and we can ignore most of it here. The only thing that still could be relevant in the above snippet, however, should be the call to  `update_fvwm_monitor(t)`, right?

But that is done identically also at the very bottom of function `monitor_update_ewmh`:

```c
for (t = Scr.FvwmRoot.next; t; t = t->next) {
  update_fvwm_monitor(t);
}
```

Wouldn't that suffice?

ThomasAdam:
Hi @farblos 

Yes. This should suffice.

farblos:
Hi @ThomasAdam,

this is it.  Hopefully.  My changes + your cleanup + your commit message - fixme, squashed and rebased onto main.  Should be identical to `ta/review-gh-1069` (after rebasing that as well) except for the fixme.

So if you also think that this is it, feel free to merge.

Thanks a lot!

P.S. I have tried to reach you on IRC wrt. some follow-up issues, will try to pester you later with these...

PR #1071, opened 2024-10-16
default-config: Don't close terminal if manpage not found.

somiaj:
Only show the "Help" menu if the manual pages are installed. This is done by testing for the fvwm3all manual page, and the item entry to popup the "Help" menu is only added to the root menu if the fvwm3all manual page exists.

### DIFF
--- a/default-config/config
+++ b/default-config/config
@@ -109,11 +109,14 @@ AddToFunc MoveToCurrent
 # Function: ViewManPage $0
 #
 # This function loads the man page $0 in an terminal
-# and is used with the help menu.
+# and is used with the help menu. If the manual page is
+# not found, point users at fvwm.org and keep terminal open.
 DestroyFunc ViewManPage
 AddToFunc   ViewManPage
 + I Exec exec $[infostore.terminal] -g 80x40 \
-  -n "Manual Page - $0" -T "Manual Page - $0" -e man "$0"
+  -n "Manual Page - $0" -T "Manual Page - $0" -e 'man "$0" || \
+  (echo "$[gt.Documentation can be found at https://www.fvwm.org]"; \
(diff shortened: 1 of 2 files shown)


somiaj:
Updated the approach here. Instead of testing for the manual page, if opening them fails, keep the terminal open using `read tmp` which is POSIX. This will keep the shell open until a user hits enter or closes the terminal. Also print a message about documentation can be found at fvwm.org.

default-config/config
```diff
@@ -109,11 +109,14 @@ AddToFunc MoveToCurrent
...
 # This function loads the man page $0 in an terminal
-# and is used with the help menu.
+# and is used with the help menu. If the manual page is
+# not found, point users at fvwm.org and keep terminal open.
 DestroyFunc ViewManPage
 AddToFunc   ViewManPage
 + I Exec exec $[infostore.terminal] -g 80x40 \
-  -n "Manual Page - $0" -T "Manual Page - $0" -e man "$0"
+  -n "Manual Page - $0" -T "Manual Page - $0" -e 'man "$0" || \
+  (echo "$[gt.Documenation can be found at https://www.fvwm.org]"; \
```
ThomasAdam, line 118: Does this work as intended?  Shouldn't it be:

```
$[gt.Documenation] can be found at https://www.fvwm.org
```

Also, I'm presuming you meant `gt.Documentation` and not `gt.Documenation`?

somiaj: I want to translate the whole string, in my tests the string returns just fine, I don't have any translations to test on that side. Though I guess there is a spelling error there. I'll update the pot file too.

Note, probably pointless here since all the documentation is not translated. I just added that in case some translator wanted to translate the default-config.

PR #1084, opened 2024-10-25
Update editorconfig for EOF Newlines

Kangie:
Also make all (non-binary) files meet the POSIX definition of files again.


Kangie:
In case we need to run this again:

`git ls-files -z | while IFS= read -rd '' f; do if file --brief --mime-encoding "$f" | grep -qv binary; then tail -c1 < "$f" | read -r _ || echo >> "$f"; fi; done`

PR #1089, opened 2024-10-28
ci: add OSX and FreeBSD

Kangie:
* **What does this PR do?**

This PR adds MacOS 13 (OSX) and FreeBSD-14 to the GitHub Actions CI


ThomasAdam:
@Kangie 

Thanks. I've used both of these "techniques" in the past and have found them to be unreliable.

I already have a separate FreeBSD runner for this which I'll end up using.

As for MacOS, well... there's no hope. All manual for now.

Thanks for keeping this a separate PR for the future if we change our mind.

PR #1094, opened 2024-10-29
PrintInfo: bindings: reduce duplication

ThomasAdam:
Simplify the flag checking for Key/Mouse bindings.


somiaj:
Tested this works. Though there are two tabs between the key/mouse name and the context.

```
[1730202172.267249] print_bindings:     Key     Up                      TWFS    CM      ShuffleDir up
```

ThomasAdam:
That's fine with me -- it's because of the missing (window_name) binding.

It's only meant for debug purposes.

PR #1096, opened 2024-10-29
IconMan: suppress `-Wformat-overflow=`

Kangie:
The compiler is unable to determine if these values are never null. Although it's unlikely to be encountered, provide fallback values in the call to ConsoleDebug that uses them directly.

### DIFF
--- a/modules/FvwmIconMan/winlist.c
+++ b/modules/FvwmIconMan/winlist.c
@@ -113,8 +113,11 @@ static int matches_string (NameType type, char *pattern, char *tname,
 
   ConsoleDebug (WINLIST, "matches_string: type: 0x%x pattern: %s\n",
 		type, pattern);
-  ConsoleDebug (WINLIST, "\tstrings: %s:%s %s:%s\n", tname, iname,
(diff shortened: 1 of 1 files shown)


Kangie:
Just for reference, the warning in question 

```
../modules/FvwmIconMan/winlist.c: In function ‘figure_win_manager’:
../modules/FvwmIconMan/winlist.c:116:38: warning: ‘%s’ directive argument is null [-Wformat-overflow=]
  116 |   ConsoleDebug (WINLIST, "\tstrings: %s:%s %s:%s\n", tname, iname,
      |                                      ^~
../modules/FvwmIconMan/winlist.c:116:38: warning: ‘%s’ directive argument is null [-Wformat-overflow=]
```
`gcc version 14.2.1 20240921 (Gentoo 14.2.1_p20240921 p1)`

eli-schwartz:
> Just for reference, the warning in question. Seems a bit confused - `figure_win_manager` is nowhere near line 116...

But the other functions in question are static, and -O3 does what?

PR #1099, opened 2024-11-02
Make xkbcommon required in meson. 

somiaj:
Noticed there were a few missing optional dependencies.

### DIFF
--- a/dev-docs/INSTALL.md
+++ b/dev-docs/INSTALL.md
@@ -29,6 +29,7 @@ system in use.
 * libx11-dev
 * libxext-dev
 * libxft-dev
+* libxkbcommon-dev
 * libxrandr-dev (>= 1.5)
 * libxrender-dev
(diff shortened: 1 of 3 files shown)


ThomasAdam:
@somiaj 

Sure. But, I would argue that xkbcommon should now be a core dependency. 

Just for meson. Although that's already implied with xfixes being introduced.

ThomasAdam:
@somiaj 

Re xkbcommon, wing were also need to change meson.build?

somiaj:
Ahh I thought you already updated that. Shouldn't we also remove the fallback code as well (though I guess this will only affect meson).

ThomasAdam:
@somiaj 

I'm not updating. Just pointing this out. I guess this is delegation by subterfuge. 🙂

I don't think we need to worry about fallback code. Just meson.

Kangie:
LGTM

PR #1113, opened 2024-11-18
Convert InfoStore to TAILQ.

### LINKED ISSUE #993: InfoStore: rewrite internals to use TAILQ instead of a linked list.
When InfoStore was written, TAILQ wasn't being used.  We should change its implementation over to use TAILQ which will reduce some of the overhead of a heap-allocated structure.

somiaj:
Use TAILQ instead of a linked list for InfoStore. This also reduces looping over the list multiple times in some situations.

This fixes #993.

### DIFF
--- a/fvwm/infostore.c
+++ b/fvwm/infostore.c
@@ -37,8 +37,6 @@
 
 /* ---------------------------- local types -------------------------------- */
 
-static MetaInfo *mi_store;
-
 /* ---------------------------- forward declarations ----------------------- */
 
 static void delete_metainfo(const char *);
@@ -47,23 +45,19 @@ static void delete_metainfo(const char *);
 
 /* ---------------------------- exported variables (globals) --------------- */
 
+struct meta_infos meta_info_q;
+
 /* ---------------------------- local functions ---------------------------- */
-
-MetaInfo *new_metainfo(void)
-{
-	MetaInfo *mi;
-
-	mi = fxcalloc(1, sizeof *mi);
-
-	return mi;
-}
 
 void insert_metainfo(char *key, char *value)
 {
 	MetaInfo *mi;
 	MetaInfo *mi_new;
 
-	for (mi = mi_store; mi; mi = mi->next)
+	if (TAILQ_EMPTY(&meta_info_q))
+		TAILQ_INIT(&meta_info_q);
+
+	TAILQ_FOREACH(mi, &meta_info_q, entry)
 	{
 		if (StrEquals(mi->key, key))
 		{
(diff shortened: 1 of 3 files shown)


ThomasAdam:
Hi @somiaj 

I think this looks good overall.  I would apply the following diff on top though:

```
diff --git a/fvwm/infostore.c b/fvwm/infostore.c
index 07cb5aafd..315b33127 100644
--- a/fvwm/infostore.c
+++ b/fvwm/infostore.c
@@ -83,7 +83,7 @@ void insert_metainfo(char *key, char *value)
 
 static void delete_metainfo(const char *key)
 {
-       MetaInfo *mi;
+       MetaInfo *mi = NULL;
 
        TAILQ_FOREACH(mi, &meta_info_q, entry)
        {
@@ -119,22 +119,17 @@ inline char *get_metainfo_value(const char *key)
 void print_infostore(void)
 {
        MetaInfo *mi;
-       bool not_found = 1;
-
-       fvwm_debug(__func__, "Current items in infostore (key, value):\n\n");
-       TAILQ_FOREACH(mi, &meta_info_q, entry)
-       {
-               fvwm_debug(__func__, "%s\t%s\n", mi->key, mi->value);
-               not_found = 0;
-       }
 
-       if (not_found)
-       {
+       if (TAILQ_EMPTY(&meta_info_q)) {
                fvwm_debug(__func__,
                           "No items are currently stored in the infostore.\n");
                return;
        }
 
+       fvwm_debug(__func__, "Current items in infostore (key, value):\n\n");
+       TAILQ_FOREACH(mi, &meta_info_q, entry)
+               fvwm_debug(__func__, "%s\t%s", mi->key, mi->value);
+
        return;
 }
```

fvwm/infostore.c
```diff
@@ -47,23 +45,19 @@ static void delete_metainfo(const char *);
...
+/* ---------------------------- local functions ---------------------------- */
 
 void insert_metainfo(char *key, char *value)
 {
 	MetaInfo *mi;
 	MetaInfo *mi_new;
 
-	for (mi = mi_store; mi; mi = mi->next)
+	if (TAILQ_EMPTY(&meta_info_q))
+		TAILQ_INIT(&meta_info_q);
```
somiaj, line 58: @ThomasAdam I am a bit concerned about this line, what if something is added to the store but then removed and it is emptied, this means the `TAILQ_INIT` is run more than once and I'm unsure if this will cause memory leaks or not. I am considering adding a bool variable `do_initialize_store` to store the state of if it has been initialized or not to keep it from being initialized more than once, but unsure if that is needed or not.

somiaj:
Patched applied, thanks.

PR #1115, opened 2024-11-19
infostore: move non-error case out of error block

ThomasAdam:
Don't hide the non-error case of adding the string to the InfoStore if
we're also suggesting the goto block is an error.

### DIFF
--- a/fvwm/infostore.c
+++ b/fvwm/infostore.c
@@ -150,13 +150,9 @@ void CMD_InfoStoreAdd(F_CMD_ARGS)
 error:
 	if (key == NULL || value == NULL) {
 		fvwm_debug(__func__, "Bad arguments given.");
-		goto out;
 	}
-out:
 	free(key);
 	free(value);
-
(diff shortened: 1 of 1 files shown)


fvwm/infostore.c
```diff
@@ -145,12 +145,13 @@ void CMD_InfoStoreAdd(F_CMD_ARGS)
...
 
+	insert_metainfo(key, value);
+
 error:
 	if (key == NULL || value == NULL) {
 		fvwm_debug(__func__, "Bad arguments given.");
 		goto out;
 	}
-	insert_metainfo(key, value);
 out:
```
somiaj, line 155: Shouldn't this just be

```
error:
	if (key == NULL || value == NULL)
		fvwm_debug(__func__, "Bad arguments given.");
```

now, no need for the `out:` label.

ThomasAdam: We can just fall-through now, yes.

PR #1131, opened 2024-12-02
Replace deprecated AM_CONFIG_HEADER with AC_CONFIG_HEADERS

lemenkov:
Replace long-time deprecated AM_CONFIG_HEADER with recommended AC_CONFIG_HEADERS macro.   AM_CONFIG_HEADER was [deprecated in 2002 and completely removed in 2012](https://lists.gnu.org/archive/html/automake/2012-12/msg00038.html).


somiaj:
Thanks for the commit, but it appears you missed the 1.1.1 release notes. Fvwm3 now uses meson as its primary build system, and autotools will be removed in 6 months. Changes or updates to autotools are no longer being accepted.

PR #1132, opened 2024-12-02
Fixes for C99 compatibility

lemenkov:
Do not check that the compiler supports implicit ints without errors.  See fvwmorg/fvwm#100.

Related to:

 * https://fedoraproject.org/wiki/Changes/PortingToModernC
 * https://fedoraproject.org/wiki/Toolchain/PortingToModernC


lemenkov:
> Hi @lemenkov
> 
> Thanks, but we are removing autotools soon, and I am not interested in accepting patches to this.
> 
> I would start investigating moving to meson.

Ok, fair enough.

PR #1142, opened 2024-12-14
ta/xfixes cursor barrier

ThomasAdam:
- **cursorbarrier: forward-declare functions**
- **cursorbarrier: destroy barriers if MAX_BARRIERS is reached**


fvwm/cursor.c
```diff
@@ -112,6 +118,7 @@ void destroy_barrier_n(int n)
 #define create_barrier(a, b, c, d)
 #define destroy_all_barriers()
 #define destroy_barrier_n(a)
+#define destroy_barrier_n(a)
```
somiaj, line 121: It appears you are defining this function again, and I'm unsure why.

fvwm/cursor.c
```diff
@@ -50,7 +55,8 @@ int num_barriers = 0;
 void create_barrier(int x1, int y1, int x2, int y2)
 {
 	if (num_barriers >= MAX_BARRIERS) {
-		fvwm_debug(__func__, "Too many barriers. Aborting.");
+		fvwm_debug(__func__, "Too many barriers. Freeing previous...");
+		destroy_all_barriers();
```
somiaj, line 59: What is your reasoning here? If they reach max barriers do you just want a reset? Since you are destroying all of the barriers first, should you remove the return bellow this so the current one they are trying to create gets created?

somiaj: You might want to document this behavior in the manual page.

ThomasAdam: It's odd either way.

No one's going to reach this limit.

Issue #1146, opened 2024-12-20
Iconized FvwmPager can't change to Page with virtual x > 32768, non Iconized FvwmPager


ileGITimo:
fvwm3 1.1.2 (80e60179)
Dual 1920x1080 screens, side by side, making up a 3840x1080 fvwm root.
DeskTopSize    12x1
On an iconized pager I can't click on rightmost pages 10 and 11, where the x/left coordinate is larger than 32768.
On a non-iconized pager it works!!
Log shows "GotoPage: invalid arguments: screen DisplayPort-0 11" or  DisplayPort-0 10.  (Screens are on DisplayPort-0 and HDMI-A-0)
 See attached video. When mouse moves over a page, I'm clicking on the mouse to move there.

https://github.com/user-attachments/assets/24e977b5-7b1b-4046-9c08-d9c8f6e95595

PR #1150, opened 2024-12-22
In IconSwitchPage char command[34] is not enough

ileGITimo:
In IconSwitchPage char command[34] is not enough, some commands gettng truncated with long screen names, increasing to 64

Fixes issue #1146

(and thus as @somiaj claimed it is not an int16 truncation problem, it is a command truncation problem)


ThomasAdam:
Thanks, but I think we can do slightly better.  See: #1151

PR #1156, opened 2024-12-29
Correct shadows on vertical titlebars

hak-foo:
* **What does this PR do?**
The "relieved rectangle with rotation" code would highlight the bottom side of the rectangle, and shade the top when a 90 or 270 degree rotation was specified.  This causes titlebars to be inconsistently highlighted.
comparison


ThomasAdam:
Hey @hak-foo 

Thanks -- this looks OK to me.  Just one thing -- please squash your two commits together; the comment correction is too minor to warrant a commit of its own in this case.

hak-foo:
Squashed commits. :)

PR #1170, opened 2025-02-13
FvwmRearrange: Use -on_screen to tile/cascade all windows on the monitor.

pghvlaans:
This PR adds a new option, `-on_screen`, to `FvwmRearrange`. It causes all windows on the monitor to be tiled or cascaded in the bounding box, even if they do not intersect it. Since it is already possible to do this on a per-desk basis with `-desk`, I thought it would be nice to have something similar for ~~pages~~ the current monitor.

### DIFF
--- a/doc/FvwmRearrange.adoc
+++ b/doc/FvwmRearrange.adoc
@@ -265,6 +265,10 @@ or cascaded.
 -desk::
   Causes all windows on the desk to be cascaded/tiled instead of only
   windows that intersect the bounding box.
+-on_screen::
+  Causes all windows on the monitor to be cascaded/tiled instead of only
+  windows that intersect the bounding box. Has no effect when used together
+  with _-desk_.
 
 === ORDERING OPTIONS
 
--- a/modules/FvwmRearrange/FvwmRearrange.c
+++ b/modules/FvwmRearrange/FvwmRearrange.c
@@ -129,6 +129,7 @@ int transients	 = 0;
 int maximized	 = 1;
 int titled	 = 1;
 int desk	 = 0;
+int on_screen	 = 0;
 int reversed	 = 0;
 int do_ewmhiwa	 = 0;
 int is_init	 = 1;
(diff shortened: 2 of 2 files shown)


doc/FvwmRearrange.adoc
```diff
@@ -265,6 +265,10 @@ or cascaded.
 -desk::
   Causes all windows on the desk to be cascaded/tiled instead of only
   windows that intersect the bounding box.
+-page::
+  Causes all windows on the monitor to be cascaded/tiled instead of only
+  windows that intersect the bounding box. Has no effect when used together
+  with _-desk_.
```
somiaj, line 271: Unsure if `-page` is a good name for this, because it only finds windows on the current monitor, which won't be all windows on a page in a multi monitor setup.

modules/FvwmRearrange/FvwmRearrange.c
```diff
@@ -810,6 +820,12 @@ update_sizes(char *argv[])
 {
 	int left = 0, top = 0;
 
+	/* Preserve unadjusted box size in case of -page */
+	pg_x = box_x;
+	pg_y = box_y;
+	pg_w = box_w;
+	pg_h = box_h;
+
```
somiaj, line 828: This really isn't unadjusted, it already includes any base struts from fvwm or window hints. If you want the full monitor size you don't need to save extra variables, just use `mon->si->{x,y,w,h}`. In this case no need to save extra variables for this.

somiaj (dismissed):

doc/FvwmRearrange.adoc
pghvlaans: You're right. I'll change the name.

modules/FvwmRearrange/FvwmRearrange.c
pghvlaans: Good catch, thank you.

pghvlaans:
Should I squash the two commits?

ThomasAdam:
> Should I squash the two commits?

Yes please.

somiaj:
I can't think of a better name than `-monitor`, but now `-screen` and `-monitor` might be confusing. Anyways I'll let @ThomasAdam take things from here.

pghvlaans:
Maybe `-from_whole_screen` or `-all_on_screen` would be clearer.

somiaj:
I like `-all_on_screen`, could probably just shorten it to `-on_screen`, making it sound more like a toggle.

ThomasAdam:
> I like `-all_on_screen`, could probably just shorten it to `-on_screen`, making it sound more like a toggle.

Agreed. `-on_screen` works for me!

PR #1171, opened 2025-02-14
SVG: Always set SVG dimentions

### LINKED ISSUE #1168: SVG icons no longer work after updating to FVWM3 1.1.2
After updating FVWM3 to version 1.1.2, SVG icons stopped working in my configuration. Previously, the following settings worked correctly:  

```
ButtonStyle 1 Pixmap thunder.svg  -- Flat  
ButtonStyle 2 Pixmap window-close-symbolic.svg -- Flat  
ButtonStyle 4 Pixmap window-maximize-symbolic.svg -- Flat  
ButtonStyle 6 Pixmap window-minimize-symbolic.svg -- Flat  
```  

However, after the update, the icons are not displayed at all.

utopiabound:
* **What does this PR do?**

This always sets a value for SVG size which accounts for SVG files that do not have a viewport.

Fixes #1168 

Without Patch (showing SVG's without `viewport` elements):
screen

With Patch (Working as well as 1.1.1 did with the same SVGs):
with-patch

### DIFF
--- a/libs/PictureImageLoader.c
+++ b/libs/PictureImageLoader.c
@@ -301,18 +301,20 @@ Bool PImageLoadSvg(FIMAGE_CMD_ARGS)
 	rsvg_handle_get_intrinsic_dimensions(rsvg, &has_out_w, &out_w,
 		&has_out_h, &out_h, &has_vb, &viewbox);
 
-	if (!has_vb) {
-		fvwm_debug(__func__, "Couldn't determine viewbox");
-		return False;
+	if (has_vb) {
+		dim.width = viewbox.width;
+		dim.height = viewbox.height;
 	}
-
-	dim.width = viewbox.width;
-	dim.height = viewbox.height;
 
 	if (has_out_w)
 		dim.em = out_w.length <= dim.width ? dim.width : out_w.length;
+	else
+		dim.em = dim.width;
+
 	if (has_out_h)
 		dim.ex = out_h.length <= dim.height ? dim.height : out_h.length;
+	else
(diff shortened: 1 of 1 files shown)


ThomasAdam:
Hi @utopiabound 

Not sure about this.  This was my original approach to handling this, and it broke SVG rendering for some versions of librsvg.

I'm inclined to just leave things as-is, and allow people to correctly "update" their SVG icons the way they're probably supposed to.

7185:
Thank you for this patch.
`viewBox` is not mandatory on SVG files, it's perfectly fine to write  `<svg width="64" height="64" ...>` for a 64 px (default unit) squared viewport even though [it's not recommended](https://css-tricks.com/scale-svg/#aa-the-svg-scaling-toolbox). These values can be overwritten if using something like `Icon file.svg:16x16`. Scaling would be different without a `viewBox` but it should still render something valid.

ExecutorElassus:
I have icons for ButtonsBar specified in ~/.fvwm/modules to use the "Icon {path/to/icon.svg}:32x32" syntax, and I still see any icon missing `viewBox` in the .svg file not getting rendered. Is there something else I'm missing here?

somiaj:
Currently you have to add the `viewBox` to the missing svgs to get them to work.

ExecutorElassus:
Yup, I went through and added that manually. Unfortunately, this is a set of icons from a github repository, so if I ever update it again I'll have to go back and do it over. But fortunately, this is only like eight icons I need to fix, so it's an easy workaround for the time being.

PR #1176, opened 2025-02-21
expand.c: Add "maximum" and "minimum" to math.

pghvlaans:
This commit would add `<` (minimum) and `>` (maximum) to the list of available operations in `$[math]`. I think this could potentially be used to do some interesting things, such as determining the orientation of a rotating monitor in a portable way.

### DIFF
--- a/doc/fvwm3_manpage_source.adoc
+++ b/doc/fvwm3_manpage_source.adoc
@@ -1637,11 +1637,12 @@ $[bg.cs3.lighten15.hash].
 +
 Please refer to the *Colorsets* section for details about colorsets.
 
-$[math.+.<x>,<y>] $[math.-.<x>,<y>] $[math.*.<x>,<y>] $[math./.<x>,<y>] $[math.%.<x>,<y>] $[math.^.<x>,<y>]::
+$[math.+.<x>,<y>] $[math.-.<x>,<y>] $[math.*.<x>,<y>] $[math./.<x>,<y>] $[math.%.<x>,<y>] $[math.^.<x>,<y>] $[math.<.<x>,<y>] $[math.>.<x>,<y>]::
 	The math expansion variables can be used to do some basic arithmetic
 	on the integers <x> and <y>. These expressions can be used to add
 	({plus}), subtract (-), multiply ({asterisk}), divide (/), modulus
-	(%), and exponentiation ({caret}). This can be useful when computing
+	(%), and exponentiation ({caret}). Maximum (>) returns the greater of
+	two values, and minimum (<) the lesser. This can be useful when computing
 	the size of panels or locations of windows relative to a monitor, for
 	example $[math.-.$[monitor.$[monitor.primary].height],200] will
 	subtract 200 pixels from the height of the primary monitor. These can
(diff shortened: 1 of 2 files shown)


somiaj:
I would like to know more about the use case, since expansion variables aren't really conditionals. To me this sounds more like conditionals should be extended to be able to test this, but unsure if expansion variables are the best place to add some conditional like this.

somiaj:
Also as a `[$math.]` operation, it seems more reasonable to me that it should return the bigger or smaller of the two values, not just a truth value.

pghvlaans:
> I would like to know more about the use case, since expansion variables aren't really conditionals. To me this sounds more like conditionals should be extended to be able to test this, but unsure if expansion variables are the best place to add some conditional like this.

It's currently possible to find out which of two values is greater indirectly by using `$[math./.x,y]` and seeing if the result is `EnvMatch` with 0. Maybe someone wants to show a program's MiniIcon in Title, but only if the window's width is over, say, 200. Of course, this approach will stop working if `math` uses floating values in a future release.

Putting < and > in `math` struck me as the most logical choice since they compare two numbers.

> Also as a `[$math.]` operation, it seems more reasonable to me that it should return the bigger or smaller of the two values, not just a truth value.

Maybe a separate expansion variable for numerical comparisons? Maximum and minimum could go in as well.

ThomasAdam:
I think we need to be a bit careful here...

The DSL in fvwm isn't formal; we're in danger here of creating a reverse-polish engine with `$[math.X.Y]` without realising it.

The original use-case of introducing these placeholders was to allow for a more direct route to performing simple arithmetic logic, without requiring things like `PipeRead` to do it.

But there will also come a point where one needs to consider `PipeRead` anyway.

Whether we allow this PR nor not, is unclear to me.

I do agree with @somiaj though that any `gt` or `lt` proposals should return the **number** and not be a boolean operation.

pghvlaans:
> Sorry, @pghvlaans -- but if I decide I wish to increase the scope of this at the expense of this PR, I shall...

Oh, sure, no worries.

pghvlaans:
Forgot to do the discussed change to maximum and minimum.

doc/fvwm3_manpage_source.adoc
```diff
@@ -1650,6 +1650,7 @@ $[math.+.<x>,<y>] $[math.-.<x>,<y>] $[math.*.<x>,<y>] $[math./.<x>,<y>] $[math.%
 	is replaced with 21.
 +
 The comma in "<x>,<y>" is to allow for a future possibility of using floats.
+Maximum (>) returns the greater of two values, and minimum (<) the lesser.
 All operations use C integer mathematics. The result of integer division is
```
somiaj, line 1654: Move this up to the list of all the other operations in the previous pargraph.  The statement is out of place here in talking about floats and integer math.

somiaj:
Your merge/force push broke your patch, it now contains all the recent stuff added to main too. The diff should only contain your changes (though unsure where the issue is I'm just going off of what github shows me the commit diff is)

ThomasAdam:
LGTM.  I'll wait for @somiaj to confirm.

@somiaj -- if you're happy with this PR, don't wait for me, please press the "Rebase and merge" button...

PR #1186, opened 2025-04-12
meson: set -liconv implicitly when iconv found

ThomasAdam:
If iconv is found, make sure that it can be linked against, by
explicitly setting -liconv; there is no pkg-config option for it on some
systems.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -273,6 +273,20 @@ endif
 iconv = dependency('iconv', required: get_option('iconv'))
 if iconv.found()
     conf.set10('HAVE_ICONV', true)
+
+    # Look to see if -liconv can be used; on some platforms where iconv
+    # is installed, -liconv might beed to be supplied; but isn't always
+    # supported (Archinux, for example).
+    #
+    # Note that the point of this test program isn't to be iconv-specific
+    # but to be sufficient enough to pass -liconv to test its existence.
+    linker_iconv_code = 'void a(void) {}; void _start(void) { a(); }'
+    linker_iconv_args = ['-lc', '-Wl']
+    if cc.links(linker_iconv_code,
+                    name : 'linker supports -liconv',
+                    args : [linker_iconv_args, '-liconv'])
+            add_project_link_arguments('-liconv', language: 'c')
+    endif
     all_found_deps += iconv
 endif
 


meson.build
```diff
@@ -273,6 +273,20 @@ endif
...
+    #
+    # Note that the point of this test program isn't to be iconv-specific
+    # but to be sufficient enough to pass -liconv to test its existence.
+    linker_iconv_code = 'void a(void) {}; void _start(void) { a(); }'
+    linker_iconv_args = ['-lc', '-Wl']
+    if cc.links(linker_iconv_code,
+                    name : 'linker supports -liconv',
+                    args : [linker_iconv_args, '-liconv'])
+            add_project_link_arguments('-liconv', language: 'c')
+    endif
```
eli-schwartz, line 289: I do not understand any of this.

Meson's builtin `dependency('iconv')` attempts to search two different providers of iconv:

- builtin, checks if this code compiles and links using only -lc:
  ```c
  #include <iconv.h>
  
  int main() {
      iconv_open("","");
  }
  ``
- system, checks if -liconv exists and `#include <iconv.h>` exists

The former case ***cannot and will not*** successfully compile and link if -liconv exists, since if -liconv exists then iconv.h will `#define iconv_open libiconv_open` and linking will fail unless you provide `libiconv_open`. And libc only provides `iconv_open`, but `-liconv` provides `libiconv_open`.

eli-schwartz (changes requested):

meson.build
eli-schwartz: Since this block already occurs inside `if iconv.found()`, it stands to reason that if `-liconv` exists, the iconv dependency already contains it. But, you could double check:

```meson
message('found iconv dependency using method: ' + iconv.type_name())
```

It will print either "builtin" or "system", and in the latter case it comes with `-liconv`.

eli-schwartz: Where precisely does one get `-liconv` from? It is only relevant to systems NOT using GNU glibc. Arch Linux does not support non-glibc installation.

thesamesam: I think the issue is at https://github.com/fvwmorg/fvwm3/blob/4592849dc38e6e96ff8feded3117e8f6f207aa68/meson.build#L129. If `libunistring` is installed, that check will fire and set `HAVE_LIBCHARSET`, which will lead to `locale_charset` being used, but nothing specifies we should link against libunistring.

ThomasAdam: > I think the issue is at
> 
> https://github.com/fvwmorg/fvwm3/blob/4592849dc38e6e96ff8feded3117e8f6f207aa68/meson.build#L129
> 
> . If `libunistring` is installed, that check will fire and set `HAVE_LIBCHARSET`, which will lead to `locale_charset` being used, but nothing specifies we should link against libunistring.

Hey @thesamesam -- thanks for your comment here.

I looked into this, and outside of ArchLinux, also installed `libunistring`.  The issue **did not** occur, but still seems present on ArchLinux.

So my concern is what's best to fix this.  Forgive me -- either your original premise here is incorrect -- or, it's ArchLinux speciic somehow.

Either way, we clearly need a solution which solves both of these.

I'll happily take any suggestions.  :)

bhaible: The `locale_charset` function exists because
1. not all systems have `nl_langinfo (CODESET)`, cf. https://www.gnu.org/software/gnulib/manual/html_node/nl_005flanginfo.html
2. on many systems, the values returned are not the same as on GNU systems. In order to avoid having to handle aliases in various places, it is the function `locale_charset` which maps system-dependent encoding names to GNU canonical encoding names (e.g. "ISO8859-1" → "ISO-8859-1", "UTF8" → "UTF-8").

`locale_charset` is part of the GNU libunistring API, https://www.gnu.org/software/libunistring/manual/html_node/uniconv_002eh.html, but nothing in fvwm3 references libunistring, therefore there is no need to talk about it here.

The only other library that installs a header file that declares `locale_charset` is GNU libiconv. It does so in two header files, `libcharset.h` and `localcharset.h`. fvwm3 includes `libcharset.h` before using `locale_charset ()`, which is fine.

GNU libiconv, however, is never supposed to be installed on GNU systems (such as Arch Linux or Gentoo with glibc). Therefore, on GNU systems, instead of asking "where can we get `locale_charset` from", the question should be "why is it attempting to use `locale_charset`? it should be using `nl_langinfo (CODESET)` instead".

In other words, fvwm3 should use `locale_charset` **if and only if** GNU libiconv has been found and is usable (i.e. you can link with it).

eli-schwartz: https://github.com/fvwmorg/fvwm3/blob/6370adf1cb3f3c1462bf7be123e1783df5ce08b3/libs/FlocaleCharset.c#L605C43-L616
https://github.com/fvwmorg/fvwm3/blob/6370adf1cb3f3c1462bf7be123e1783df5ce08b3/meson.build#L147
https://github.com/fvwmorg/fvwm3/blob/6370adf1cb3f3c1462bf7be123e1783df5ce08b3/libs/FlocaleCharset.h#L14-L24

It already is kinda doing that, but confusingly tries using header presence as a sentinel for "not GNU libc" and confuses it with implicit -liconv.

ThomasAdam:
Thanks all.  Spoken with @Kangie who has provided PR #1203 -- I've tested this on a fair few systems now and this appears to work fine for me.

Thanks for all your help with this.

PR #1189, opened 2025-05-03
FvwmPager: correct fonts and balloons initialization

ONykyf:
Prevent some weird FvwmPager crashes and balloons not showing up because fonts and boolean flags are not initialized an start. Also restore the functionallity of "Balloons" option without parameters.

Sometimes (and almost always when swallowed by FvwmButtons) FvwmPager silently crashes. Debugging has shown that this happens in do_label_window() function because Scr.winFfont is uninitialized. Also balloons are not always shown in FvwmPager because Balloon.show_in_pager and Balloon.show_in_icon boolean flags are undetermined, and
Balloons with no parameters has no effect. This has been corrected.

### DIFF
--- a/modules/FvwmPager/init_pager.c
+++ b/modules/FvwmPager/init_pager.c
@@ -246,8 +246,14 @@ void initialize_fonts(void)
 	/* Initialize fonts. */
 	FlocaleInit(LC_CTYPE, "", "", "FvwmPager");
 
-	/* load a default font. */
-	Scr.Ffont = FlocaleLoadFont(dpy, NULL, MyName);
+	/* load default fonts. */
+	Scr.Ffont = NULL;
+	Scr.winFfont = NULL;
+	Balloon.Ffont = NULL;
+
+	/* balloons are not shown by default */
+	Balloon.show_in_pager = false;
+	Balloon.show_in_icon = false;
 
 	/* init our Flocale window string */
 	FlocaleAllocateWinString(&FwinString);
@@ -372,6 +378,12 @@ void initialize_pager_size(void)
 	struct fpmonitor *fp = fpmonitor_this(NULL);
 	int VxPages = fp->virtual_scr.VxPages;
 	int VyPages = fp->virtual_scr.VyPages;
+
+	if (use_window_label && !Scr.winFfont)
+		Scr.winFfont = FlocaleLoadFont(dpy, NULL, MyName);
+
+	if ((use_desk_label || use_monitor_label) && !Scr.Ffont)
+		Scr.Ffont = FlocaleLoadFont(dpy, NULL, MyName);
 
 	/* Grid size */
 	if (Rows < 0)
@@ -1118,12 +1130,14 @@ void parse_options(void)
 					icon_yneg = true;
 			}
 		} else if (StrEquals(resource, "Font")) {
-			if (strncasecmp(next, "none", 4) == 0) {
+			FlocaleUnloadFont(dpy, Scr.Ffont);
+			if (strncasecmp(next, "none", 4) != 0)
+				Scr.Ffont = FlocaleLoadFont(
+					dpy, next, MyName);
+			else {
+				Scr.Ffont = NULL;
 				use_desk_label = false;
 				use_monitor_label = false;
-			} else {
-				FlocaleUnloadFont(dpy, Scr.Ffont);
-				Scr.Ffont = FlocaleLoadFont(dpy, next, MyName);
 			}
 		} else if (StrEquals(resource, "Pixmap") ||
 			   StrEquals(resource, "DeskPixmap"))
@@ -1187,9 +1201,15 @@ void parse_options(void)
 			   StrEquals(resource, "SmallFont"))
 		{
 			FlocaleUnloadFont(dpy, Scr.winFfont);
-			if (strncasecmp(next, "none", 4) != 0)
+			if (strncasecmp(next, "none", 4) != 0) {
 				Scr.winFfont = FlocaleLoadFont(
 					dpy, next, MyName);
+				if (Scr.winFfont)
+					use_window_label=true;
+			} else {
+				Scr.winFfont = NULL;
+				use_window_label=false;
+			}
 		} else if (StrEquals(resource, "Rows")) {
 			sscanf(next, "%d", &Rows);
 		} else if (StrEquals(resource, "Columns")) {
@@ -1237,8 +1257,15 @@ void parse_options(void)
 				Balloon.show_in_icon = true;
 			}
 		} else if (StrEquals(resource, "BalloonFont")) {
-			FlocaleUnloadFont(dpy, Balloon.Ffont);
-			Balloon.Ffont = FlocaleLoadFont(dpy, next, MyName);
+		        FlocaleUnloadFont(dpy, Balloon.Ffont);
+			if (strncasecmp(next, "none", 4) != 0) {
+				Balloon.Ffont = FlocaleLoadFont(
+					dpy, next, MyName);
+			} else {
+				Balloon.Ffont = NULL;
+				Balloon.show_in_pager = false;
+				Balloon.show_in_icon = false;
+			}
 		} else if (StrEquals(resource, "BalloonBorderWidth")) {
 			sscanf(next, "%d", &(Balloon.border_width));
 		} else if (StrEquals(resource, "BalloonYOffset")) {
--- a/modules/FvwmPager/FvwmPager.c
+++ b/modules/FvwmPager/FvwmPager.c
@@ -96,6 +96,7 @@ bool	HilightDesks = true;
 bool	HilightLabels = true;
 bool	error_occured = false;
 bool	use_desk_label = true;
+bool	use_window_label = true;
 bool	WindowBorders3d = false;
 bool	HideSmallWindows = false;
 bool	SendCmdAfterMove = false;
--- a/modules/FvwmPager/FvwmPager.h
+++ b/modules/FvwmPager/FvwmPager.h
@@ -221,6 +221,7 @@ extern bool	HilightDesks;
 extern bool	HilightLabels;
 extern bool	error_occured;
 extern bool	use_desk_label;
+extern bool	use_window_label;
 extern bool	WindowBorders3d;
 extern bool	HideSmallWindows;
 extern bool	SendCmdAfterMove;


ThomasAdam:
Hi @ONykyf 

This might fix the immediate issue, but I don't think it is the way to go.

I would rather see the `Balloon` struct initialised in `initialize_balloon_window()` -- you might have to rejig the ordering of how that's called in `init_fvwm_pager()`.

ONykyf:
All fallback fonts initialization is in initialize_pager_window() and initialize_balloon_window() now.

ThomasAdam:
Hi @ONykyf 

Better.  Thanks!  But you need to squash those two commits together, please.

ThomasAdam:
Hi @ONykyf 

`FvwmPager` now segfaults with this change.  I'm presuming something like the following fixes it:

```
diff --git a/modules/FvwmPager/x_pager.c b/modules/FvwmPager/x_pager.c
index 7dc6819d0..c2ca3f71b 100644
--- a/modules/FvwmPager/x_pager.c
+++ b/modules/FvwmPager/x_pager.c
@@ -336,7 +336,7 @@ rectangle set_vp_size_and_loc(struct fpmonitor *m, bool is_icon)
 
 void set_desk_size(bool update_label)
 {
-       if (update_label) {
+       if (update_label && Scr.Ffont != NULL) {
                label_h = 0;
                if (use_desk_label)
                        label_h += Scr.Ffont->height + 2;
```

But you need to be very careful with the testing you do...

For reference:

```
Core was generated by `/usr/local/libexec/fvwm3/1.1.3/FvwmPager 11 5 none 0 8 * *'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  set_desk_size (update_label=true) at ../modules/FvwmPager/x_pager.c:342
342                             label_h += Scr.Ffont->height + 2;
(gdb) bt
#0  set_desk_size (update_label=true) at ../modules/FvwmPager/x_pager.c:342
#1  0x000055870ce4b641 in initialize_pager_size () at ../modules/FvwmPager/init_pager.c:410
#2  0x000055870ce4a763 in init_fvwm_pager () at ../modules/FvwmPager/init_pager.c:68
#3  0x000055870ce49f56 in main (argc=8, argv=0x7ffe1247a518) at ../modules/FvwmPager/FvwmPager.c:292
(gdb) p 2
$1 = 2
(gdb) l
337     void set_desk_size(bool update_label)
338     {
339             if (update_label) {
340                     label_h = 0;
341                     if (use_desk_label)
342                             label_h += Scr.Ffont->height + 2;
343                     if (use_monitor_label)
344                             label_h += Scr.Ffont->height + 2;
345             }
346     
(gdb) f 2
#2  0x000055870ce4a763 in init_fvwm_pager () at ../modules/FvwmPager/init_pager.c:68
68              initialize_pager_size();
(gdb) f 1
#1  0x000055870ce4b641 in initialize_pager_size () at ../modules/FvwmPager/init_pager.c:410
410             set_desk_size(true);
(gdb) quit
```

ONykyf:
Yes, this is because initialize_pager_size() comes before initialize_pager_window(), and font initialization is in the latter. I think it is better to move it to the first, then the proposed check Scr.Ffont != NULL is unnecessary, although can be retained for safety.

modules/FvwmPager/init_pager.c
```diff
@@ -688,6 +690,12 @@ void initialize_pager_window(void)
 		(NorthWestGravity)	/* gravity */
 	};
 
+	if (use_window_label && !Scr.Ffont)
+		Scr.winFfont = FlocaleLoadFont(dpy, NULL, MyName);
```
somiaj, line 694: Shouldn't this be `!Scr.winFfont` here to match what you are actually setting.

ONykyf: Thank You, it should be `!Scr.winFfont` indeed.

modules/FvwmPager/x_update.c
```diff
@@ -284,7 +284,7 @@ void update_window_decor(PagerWindow *t)
 
 void do_label_window(PagerWindow *t, Window w, rectangle r)
 {
-	if (t == NULL || w == None || Scr.winFfont == NULL ||
+	if (t == NULL || w == None || !use_window_label ||
```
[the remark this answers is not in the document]
ONykyf: Now the logic of the code is such that after initialize_pager_size() is completed, we have either use_window_label == false and Scr.winFfont == NULL, or use_window_label == true and Scr.winFfont == whatever non-zero, which is either default or non-default font.  But inside parse_options(), yes, it is safer to leave `Scr.winFfont == NULL` check as is.

ThomasAdam:
I can't say I like the churn on this change -- I'm not a fan.

We seem to have now decided on resetting `Balloon.Ffont` and `Scr.Ffont` back to NULL in several places along different code paths, which just makes this look a bit of a mess.

I would say "try again".  I'm sure you can think of a better way, @ONykyf...

ONykyf:
Maybe I have not understood the intentions behind this code well, then I'd like to ask for a clarification.

My idea was the following: first we declare what we are going and what we are going to use: 
`use_desk_label = true;
use_window_label = true;
use_monitor_label = false;`

All fonts are set to NULL == not set. Then parse_options() can consume "Font", "WindowFont", "SmallFont", or "BalloonFont" options arbitrarily many times. Each time, if the option value is "None", then the respective boolean flag is set to false (we have no intention to use this), and the font pointer to NULL. Otherwise, if the option value is not "None", we set the respective font.  We can reconsider, then the previous choice is overwritten.

Finally, if we have not made a choice, but the intention to use some font has not been rejected, then the default font is used.

Is this logic acceptable?

modules/FvwmPager/init_pager.c
```diff
@@ -953,6 +961,9 @@ void parse_options(void)
 			use_no_separators = true;
 		} else if (StrEquals(resource, "IgnoreWorkingArea")) {
 			ewmhiwa = true;
+		} else if (StrEquals(resource, "Balloons")) {
+			Balloon.show_in_pager = true;
+			Balloon.show_in_icon = true;
```
somiaj, line 966: This should not be here, this is already set on lines 1125, and correctly honors the options, this is overriding what is done below and making it so the option does not work according to the manual page.

somiaj: Oh I see what this is doing, but it is not doing it correctly. It does appear that the parser will ignore `*FvwmPager: Balloons`, without any options. It does state that this should enable balloons in the manual page, though really just use `*FvwmPager: Balloons All` or anything except null and it will process correctly. I'm not sure the best thing to do here, but this is not it. This will make it so `Balloons None` or `Balloons Pager` will fail.

I might just update the manual page to state that you have to provide an option vs try to work around dealing with it being Null.

somiaj:
Can you share a config that causes the crash you are trying to fix? That way we can test what is actually crashing. The default-config and I use balloon windows and I have never seen FvwmPager crash due to this when swallowed inside of FvwmButtons. I also think your fix is trying to add functionality we may not need, but if we better knew what was crashing we can maybe suggest a better way to fix the initialization issue (really testing of a font is Null is a sufficient test, a new variable doesn't need to be added).

Also why are you adding a new configuration option "Balloons" without any documentation on what it does?

modules/FvwmPager/init_pager.c
somiaj: Note the default config is using `*FvwmPager: Balloons All`

ONykyf: > Oh I see what this is doing, but it is not doing it correctly. It does appear that the parser will ignore `*FvwmPager: Balloons`, without any options. It does state that this should enable balloons in the manual page, though really just use `*FvwmPager: Balloons All` or anything except null and it will process correctly. I'm not sure the best thing to do here, but this is not it. This will make it so `Balloons None` or `Balloons Pager` will fail.
> 
> I might just update the manual page to state that you have to provide an option vs try to work around dealing with it being Null.

I just followed FvwmPager.adoc (and hence FvwmPager.1): "If _type_ is anything else (**or null**) balloons are always shown, so no need to change the man page." I had `*FvwmPager: Balloons` and then `*FvwmPager: BalloonFont "xft:Sans:size=8:medium:antialias=True"` in my config, and this did not worked (crashed).

ONykyf: > This should not be here, this is already set on lines 1225, and correctly honors the options, this is overriding what is done below and making it so the option does not work according to the manual page.

I would not say so, on lines 1225-1238 `Balloon.show_in_pager` and `Balloon.show_in_icon` are set only if Balloons option with (and now without) parameters is used. Otherwise these flags are never initialized. Probably this should also be corrected?

ONykyf:
> Can you share a config that causes the crash you are trying to fix? That way we can test what is actually crashing. The default-config and I use balloon windows and I have never seen FvwmPager crash due to this when swallowed inside of FvwmButtons. 

In fact FvwmPager crashed always, but immediately when swallowed by FvwmButtons, or after Midnight Commander (or any other program like xterm that changes its title while running) is started. Nothing interesting in the config, just 

*FvwmPager: UseSkipList
*FvwmPager: Back grey60
*FvwmPager: Fore MidnightBlue
*FvwmPager: SolidSeparators
*FvwmPager: Font "xft:Sans:size=8:medium:antialias=True"
*FvwmPager: Hilight #c3c3c3
*FvwmPager: Geometry 90x70-25-35
*FvwmPager: SmallFont "xft:Sans:size=7:medium:antialias=True"
*FvwmPager: Balloons
*FvwmPager: BalloonFont "xft:Sans:size=8:medium:antialias=True"

> I also think your fix is trying to add functionality we may not need, but if we better knew what was crashing we can maybe suggest a better way to fix the initialization issue (really testing of a font is Null is a sufficient test, a new variable doesn't need to be added).
> 
> Also why are you adding a new configuration option "Balloons" without any documentation on what it does?

As You can see, Balloons option is already there and does not work as announced, so no new new functionality, just aligning the code with the manual page.

New boolean variable `use_window_label` is necessary to distinguish two situations when `Scr.winFfont == NULL` : when neither `WindowFont` nor `SmallFont` has not been set, but we accept the default font, from the case when `WindowFont` or `SmallFont` has been set to "None", hence no font should be loaded at all. Once the initialization is complete, this variable is not used anymore.

ONykyf:
Correct so that now Balloons option looks for an argument (thanks to @somiaj).

Set initial values for `Balloon.show_in_pager` and `Balloon.show_in_icon` to `true` to avoid a lottery if no options are set by a user (or different defaults are preferred?)

modules/FvwmPager/init_pager.c
somiaj: Even if you added `flags = 0` to continue parsing, it would still fail because it won't continue if there is no argument. I am going to update the manual page, `*FvwmPager: Balloons` is now invalid, use `*FvwmPager: Balloons All`, the updates to the parser make it not worth adding a work around to accept `null` in this case any more.

ONykyf: > Even if you added `flags = 0` to continue parsing, it would still fail because it won't continue if there is no argument. I am going to update the manual page, `*FvwmPager: Balloons` is now invalid, use `*FvwmPager: Balloons All`, the updates to the parser make it not worth adding a work around to accept `null` in this case any more.

Could You explain why it would still fail? I there is no argument to Balloons, it will just free `resource`, `arg1` and `arg2` and skip to the next cycle iteration, but `Balloon.show_in_pager` and `Balloon.show_in_icon` have been already set to true, so everything is ok.

Of course You can remove this option, but I see no problem with it now.

somiaj: Line 990 bellow,

```
if (!arg1)
			/* No inputs, nothing to do. */
			continue;
```

If there is no arg, it will also not continue processing. All options below are expected to have an argument, which is why I said I didn't want to design a special case for this option when just not allowing null and removing the comment that it is allowed from the manual page is my preferred solution. Really this would be a single exception that really isn't needed. People can update their configuration.

ThomasAdam: > People can update their configuration.

Yep.  I'm more than happy for this -- wouldn't be the first time, and certainly will not be the last.  ;)

ONykyf: Ok, removing (null) case.

ONykyf:
@somiaj I would rather leave editing manpage to You, I am not a native English speaker.

somiaj:
I really want to track down the crash you are experiencing, because I want to ensure that the changes are needed. So far I cannot reproduce the crash, I have used your configuration and swallowed it in FvwmButtons as shown.

```
DestoryModuleConfig FP1:*
*FP1: UseSkipList
*FP1: Back grey60
*FP1: Fore MidnightBlue
*FP1: SolidSeparators
*FP1: Font "xft:Sans:size=8:medium:antialias=True"
*FP1: Hilight #c3c3c3
*FP1: Geometry 90x70-25-35
*FP1: SmallFont "xft:Sans:size=7:medium:antialias=True"
*FP1: Balloons All
*FP1: BalloonFont "xft:Sans:size=8:medium:antialias=True"

DestoryModuleConfig FPButtons:*
*FPButtons: Geometry 400x400
*FPButtons: Rows 1
*FPButtons: Columns 1
*FPButtons: (1x1, Swallow FP1 'Module FvwmPager FP1 *', Frame 0)

Module FvwmButtons FPButtons
```

I have changed the title of my `xterm` multiple times and it is not crashing for me.

somiaj:
Ahh I see now, setting the font to `None` after it has been set does cause a crash. This is due to bad configuration (no reason to set the font then not set it), but I will think about a good way to protect against that.

ONykyf:
> Ahh I see now, setting the font to `None` after it has been set does cause a crash. This is due to bad configuration (no reason to set the font then not set it), but I will think about a good way to protect against that.

That was the reason why I proposed to set NULL in this case.

somiaj:
I am able to fix that issue with a single line:

```diff
--- a/modules/FvwmPager/init_pager.c
+++ b/modules/FvwmPager/init_pager.c
@@ -1187,6 +1187,7 @@ void parse_options(void)
                           StrEquals(resource, "SmallFont"))
                {
                        FlocaleUnloadFont(dpy, Scr.winFfont);
+                       Scr.winFfont = NULL;
                        if (strncasecmp(next, "none", 4) != 0)
                                Scr.winFfont = FlocaleLoadFont(
                                        dpy, next, MyName);
```

modules/FvwmPager/init_pager.c
```diff
@@ -246,8 +246,14 @@ void initialize_fonts(void)
...
-	/* load a default font. */
-	Scr.Ffont = FlocaleLoadFont(dpy, NULL, MyName);
+	/* load default fonts. */
+	Scr.Ffont = NULL;
+	Scr.winFfont = NULL;
+	Balloon.Ffont = NULL;
+
+	/* balloons are shown by default */
+	Balloon.show_in_pager = true;
+	Balloon.show_in_icon = true;
```
somiaj, line 256: Please don't change default behavior. Balloons are not shown by default.

modules/FvwmPager/init_pager.c
```diff
@@ -487,7 +499,7 @@ void initialize_balloon_window(void)
 		Balloon.y_offset = DEFAULT_BALLOON_Y_OFFSET;
 	if (Balloon.label_format == NULL)
 		Balloon.label_format = fxstrdup("%i");
-	if (!Balloon.Ffont)
+	if ((Balloon.show_in_icon || Balloon.show_in_pager) && !Balloon.Ffont)
```
somiaj, line 502: This is pointless, `Balloon.show_in_pager` or `Balloon.show_in_icon` is not set, this code is never reached in line 484 above.

modules/FvwmPager/init_pager.c
ONykyf: > Please don't change default behavior. Balloons are not shown by default.

I have not found the place where `Balloon.show_in_pager` and `Balloon.show_in_icon` are set to `false`. I asked what are preferred inital behaviour, but it must be set.

modules/FvwmPager/init_pager.c
ONykyf: I agree, You are right.

modules/FvwmPager/init_pager.c
somiaj: Variables are false by default, so that is why they were not explicitly set to false. Current behavior is balloon windows are not shown. This has been the default for a while now, no need to change that.

ONykyf:
> The only crash or issue that I saw was with setting `SmallFont None`, setting `Font None` and even turning on `DeskLabels` afterwards had no issue. `BalloonFont None` is not and hasn't ever been valid. If you want to disable balloon windows use `Balloons None`.
> 
> I'm just trying to identify what issues are being fixed. Anyways, as I see it a single line will stop the crash with setting `SmallFont None` after it has already been set, and the manual page should be updated to clarify that `Balloons` must have an input and that `Balloons None` turns off showing the balloons.
> 
> Everything else seems to be wanting to change default behavior / add features / try to make things more unfiorm, but I don't see an issue outside of that crash. I agree that things aren't uniform, BalloonFont is treated differently than other Fonts, but this is an historical blunder I don't think we need to fix.

Probably, as a mathematician, I prefer things to be uniform:)  And I would not like a font to be loaded just to be removed. So what with initial values for `Balloon.show_in_pager` or `Balloon.show_in_icon`? Where do they become `false `?

somiaj:
Sure it would be nice if the fvwm code base was more uniform, welcome to 30 years of people doing adding features / patches / etc, also back in the last 90s and early 2000s when most this code was written, standards were far looser with code than they in modern times. Anyways I don't think this is going to be accepted as is, you are changing things that don't need to be changed, and the changes will probably lead to segfaults for other setups as was commented above. A lot more is going to have to be done to make sure that these changes are correctly adapted though the full code, and they aren't really achieving anything.

Are you having any other issue with FvwmPager not working due to explicitly setting variables to `false` initially? If so we can address that, but I have a small patch that address the actual issues you have pointed out.

I agree it would be great if the code was more uniform, but that would require a much larger change than just trying to piece in the pieces you are here.

modules/FvwmPager/init_pager.c
```diff
@@ -1118,12 +1130,14 @@ void parse_options(void)
...
+			if (strncasecmp(next, "none", 4) != 0)
+				Scr.Ffont = FlocaleLoadFont(
+					dpy, next, MyName);
+			else {
+				Scr.Ffont = NULL;
 				use_desk_label = false;
 				use_monitor_label = false;
-			} else {
-				FlocaleUnloadFont(dpy, Scr.Ffont);
-				Scr.Ffont = FlocaleLoadFont(dpy, next, MyName);
```
somiaj, line 1126: This is the reason @ThomasAdam got a segfault and will cause other segfaults. `Scr.Ffont` is expected to be a valid font in the code, so that is why it was explicitly not unloaded with the `None` setting. You are unloading the font when it shouldn't be.

ONykyf: No, this is not possible. All references to Scr.Ffont are either guarded by `use_desk_label` or `use_monitor_label`, or are inside` draw_desk_label()` all calls of which are guarded by `use_desk_label` or `use_monitor_label`, hence this cannot cause a crash. I wrote previously why @ThomasAdam got a crash.

PR #1190, opened 2025-05-03
fvwm: make rounded windows and menus corners

ONykyf:
Bresenham algorithm is used to make (some or all) windows and menus corners rounded. RoundedCorners,  RoundedCornersTop, and RoundedCornersBottom are added to Style options (with arbitrary radii), and RoundedCorners  (radius=6) and SlightlyRoundedCorners  (radius=3) are added to MenuStyle options.

* **Screenshots**
XCalc is with RoundedCornersTop (default radii), XClock with RoundedCorners 10 18 25 0, taskbar at the top with RoundedCornersBottom 8 8.
roundedcorners

MenuStyle contains RoundedCorners.
image

* **Request #946**

### DIFF
--- a/doc/fvwm3_manpage_source.adoc
+++ b/doc/fvwm3_manpage_source.adoc
@@ -2518,7 +2518,8 @@ using the *ChangeMenuStyle*. See also *DestroyMenuStyle*. When using
 monochrome color options are ignored.
 +
 _options_ is a comma separated list containing some of the keywords
-Fvwm / Mwm / Win, BorderWidth, HilightBack / !HilightBack, HilightTitleBack,
+Fvwm / Mwm / Win, BorderWidth, RoundedCorners, SlightlyRoundedCorners,
+HilightBack / !HilightBack, HilightTitleBack,
 ActiveFore / !ActiveFore, MenuColorset, ActiveColorset, GreyedColorset,
 TitleColorset, Hilight3DThick / Hilight3DThin / Hilight3DOff,
 Hilight3DThickness, Animation / !Animation, Font, TitleFont,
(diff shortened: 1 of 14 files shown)


somiaj:
Personally I won't support this until the borders are rounded correctly, and work with non-flat CDE/Motiff style decors.

ONykyf:
> Personally I won't support this until the borders are rounded correctly, and work with non-flat CDE/Motiff style decors.

The first is possible (although complicated), the second is difficult to imagine. How to handle correctly, e.g., shadows of the outermost additional frame which are of differents colors, and buttons at the corners?

somiaj:
That is just my opinion, others will see if the patch is good enough quality. But I think there is a reason that this patch hasn't been adopted by fvwm yet. I think one is it really only works in very limited situation, the more you can improve that the better chance you'll have of it being accepted.

I do notice you have `modules/FvwmPager/init_pager.c.orig` in this PR. I think that is a mistake.

ONykyf:
> That is just my opinion, others will see if the patch is good enough quality. But I think there is a reason that this patch hasn't been adopted by fvwm yet. I think one is it really only works in very limited situation, the more you can improve that the better chance you'll have of it being accepted.

I will add rounded borders, but have no idea what to do with CDE / Motif decor.
> 
> I do notice you have `modules/FvwmPager/init_pager.c.orig` in this PR. I think that is a mistake.

Thank You, it has crept somehow into the commit.

ONykyf:
> Personally I won't support this until the borders are rounded correctly, and work with non-flat CDE/Motiff style decors.

The borders of menus now look like this:
screen

What should be considered the correct rounding of borders for a window with probably all 4 border widths (in a size_borders structure) different?

somiaj:
That looks better, thanks for trying to improve the patch. I can't say too much on the actual quality of the code, we'll have to wait for @ThomasAdam to do that.

ThomasAdam:
Hi @ONykyf 

Thanks for sending this over.

I know this is a request which is often asked about -- and there's a reason I've not added it to fvwm; all attempts to do this are really invasive, and also, the implementations fail to take into account the many different decor options which are available.

Because of that, this feature is best left to when/if such a rewrite occurs -- I am sure it probably never will happen -- but neither will I cave in and add it "just because".

I'm also not in favour of the Bresenham approach as the cornerns are far too jagged; applying a smoother shapemask to these is the way to go -- which is what I think the original patch for this idea (fluxbox-inspired, ISTR) did.

Thanks anyway -- I hope if nothing else you learned a few things patching this, although it does seem to be a copy/paste job in part from other rounded corner implementations.

PR #1191, opened 2025-05-10
Remove autotools & modernise build CI/CD

ThomasAdam:
This PR addresses the change to the infrastructure of fvwm3, removing autotools and asserting meson as the primary/only build-tool to use.

This hasn't been mandatory -- it's gone through an extensive consoltation period over many months to get to this point -- as individual commits will attest to.

In addition to this change though, there is a wider implication around changing the buildsystem fvwm3 uses to make use of meson.

In doing so, this change has also meant automating fvwm3's release, removing key persons as a bottle-neck to that.


somiaj (approved):
Looks good. I didn't test the release workflow, but I did testing building a debian package, and the debian package not only builds, I can build it without specifically telling debian helper to use meson, since the autotool files are no longer present.

PR #1203, opened 2025-05-23
meson: do a link test for `locale_charset()`

Kangie:
We can't just check for `locale_charset` as some (very silly) distributions ship, or allow end users to install, GNU libiconv on GNU libc systems.

- Move `HAVE_CODESET` and `HAVE_LIBCHARSET` checks into `iconv` dep conditional
- perform link test against libcharset to see if it is usable
- add a visible warning that GNU libiconv should not be installed on a GNU libc system, if libcharset (libiconv) linking fails due to apparent conflicts with glibc iconv.

See Also: https://github.com/fvwmorg/fvwm3/pull/1186

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -126,53 +126,43 @@ foreach h : headers
     endif
 endforeach
 
-if cc.has_header_symbol('libcharset.h', 'locale_charset')
-    conf.set10('HAVE_LIBCHARSET', true)
-endif
-
 if conf.has('HAVE_SYS_SELECT_H')
     conf.set10('HAVE_SELECT', true)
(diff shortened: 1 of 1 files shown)


Kangie:
Still a little concerned about whether it should be possible to use `HAVE_CODESET` and `HAVE_LIBCHARSET` at the same time.

https://github.com/fvwmorg/fvwm3/blob/6dbc601851087c69a08ea2fa4a79b5063fe958d5/meson.build#L270-L272  

musl, for example, "has" `nl_langinfo` but libcharset will likely work there and maybe should be preferred?

PR #1206, opened 2025-05-29
fvwm: allow rounded corners for application windows and menus

ONykyf:
New features and styles are introduced, mainly to get rounded corners for window borders, buttons, menus etc. To achieve this, the code that draws relief rectangles has been heavily modified, mainly to provide correct junctions when two arms of an angle or two halves of an arc are drawn with different GCs. Bresenham algorithm is applied to construct arcs, both for drawing and for making windows shaped at corners using XShape (no antialising, hence some jagging is present, it is up to user whether to round corners or not).

Another distinctive feature is the possibility to specify different radii and corner lengths (handle lengths) for all corners of a window by `CornerLength` and `RoundedCorners` styles.

Decor elements are correctly rounded and rotated if `TitleAt{TOP,Left,Bottom,Right}` with `XXXTitleRotated{CW,CCW}` is used. An option `RotateShadows` to rotate shadows together with title text is also introduced.

screenshot-rounded

### DIFF
--- a/doc/fvwm3_manpage_source.adoc
+++ b/doc/fvwm3_manpage_source.adoc
@@ -2519,7 +2519,9 @@ using the *ChangeMenuStyle*. See also *DestroyMenuStyle*. When using
 monochrome color options are ignored.
 +
 _options_ is a comma separated list containing some of the keywords
-Fvwm / Mwm / Win, BorderWidth, HilightBack / !HilightBack, HilightTitleBack,
+Fvwm / Mwm / Win, BorderWidth, RoundedCorners / !RoundedCorners,
+SlightlyRoundedCorners / !SlightlyRoundedCorners,
+HilightBack / !HilightBack, HilightTitleBack,
 ActiveFore / !ActiveFore, MenuColorset, ActiveColorset, GreyedColorset,
 TitleColorset, Hilight3DThick / Hilight3DThin / Hilight3DOff,
 Hilight3DThickness, Animation / !Animation, Font, TitleFont,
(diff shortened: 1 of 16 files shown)


ONykyf:
Hi @ThomasAdam 

To tell the truth, I am not much surprised that this PR has been rejected. It is really invasive, and it should be such because rectangles that create relief are in many places, so many changes are introduced at once. The current method of drawing rectangles cannot be adapted to make them rounded, so a massive rewrite was necessary. Nevertheless the logic of decors creation has not been changed, so when (if) You overhaul decor code someday (I doubt that it will be soon, if will be at all), it will not pose too much difficulty to replace sharp corners with rounded ones and to add special treating for all four corners. 

This was the reason I wrote this code - I need it, it makes mu UX with FVWM more convenient and pleasant, so I will keep this patch for personal use and update it as new fvwm releases are coming - kind of a local soft fork. I do this with many projects - XOrg, Wine, SLIM etc. Some my improvements are accepted into mainstream, some are not - no problem. I see no reason to endure inconveniences if a small patch can make life better, as it was with the latest FVWM failure to process SVG images without viewport. Why should a user bother to change or replace SVGs if two lines can be changed to use width and height as a resort if no viewport is provided?

This contradicts a bit to Your fear that some people's setups will be broken. AFAICS the approach "Let people update their configuration" is used if the developers are really interested in a feature. In fact there is no other way to "check man different decor configurations to ensure it's not broken anything" than to share a code with end users that will either complain or not.

I understand that a (not very large, I guess) developers team feels more comfortable if all code is written by themselves, hence is clear and well understood. To accept large chunks of other people's code leads to risks, and everyone prefers his/her own code over other's one. Then only small bugs correction and reporting is left to outside contributors, which is a closed enough model of development. Development pace is slowed, and user base does not increase (and being visually appealing in also important here).

In this regard, don't You think about some kind of FVWM-staging like Wine-staging where users could grab and test some experimental code, and report bugs (if any)?

PR #1212, opened 2025-06-13
FvwmForm: process UTF-8 input and paste request in input fields

### LINKED ISSUE #1211: FvwmForm input field fails to process multibyte chars
In FVWM3 1.1.3, Slackware 15-current, non-latin UTF chars cannot be entered into an input field of FvwmForm - nothing happens or only one byte of a char is taken.

ReadXServer.c code looks like all characters are assumed to be one-byte.

ONykyf:
To make FvwmForm input fields UTF-8 aware, input.size field is duplicated as input.width, the first one containing size in bytes and the latter width in multibyte chars. To allow walking along a string with chars of different lengths (1 to 4 bytes), e.g., for pasting or moving a cursor by arrow keys or by a mouse click, a helper function find_nth_UTF8_char() is introduced.

Fixes #1211

### DIFF
--- a/modules/FvwmForm/FvwmForm.c
+++ b/modules/FvwmForm/FvwmForm.c
@@ -936,6 +936,11 @@ static void ct_Input(char *cp)
   while (isspace((unsigned char)*cp)) cp++;
   item->input.init_value = fxstrdup("");	    /* init */
   if (*cp == '\"') {
+    int num = -1, len;
+    char *c;
+    c = find_nth_UTF8_char(cp, NULL, &num, &len);
+    c += len;  /* the end or the first invalid UTF-8 character */
+    *c = '\0'; /* stop here */
     free(item->input.init_value);
     item->input.init_value = CopyQuotedString(++cp);
   }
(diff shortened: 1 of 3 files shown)


ONykyf:
@ThomasAdam @somiaj No more places affected by "multibyteness" found. Testing shows no problems. Please review.

ONykyf:
Hope now the logic of the helper function is more straightforward and manageable enough by whoever will do this in the future.

modules/FvwmForm/FvwmForm.c
```diff
@@ -2227,9 +2221,21 @@ char* find_nth_UTF8_char(char *str, char *before,
...
 #endif
     }
-    return NULL; /* Sentinel */
+
+    if (num && *num >= 0 && i <= *num) {
+    /* ended prematurely => return '\0' of length zero */
+	pstr = str;
+	l = 0;
+    }
+    else if ((l == 0) || (*str == '\0')) { /* invalid UTF-8 char or '\0' */
```
ThomasAdam, line 2233: We said we weren't going to do this `1 == 0` business....

modules/FvwmForm/FvwmForm.c
```diff
@@ -2227,9 +2221,21 @@ char* find_nth_UTF8_char(char *str, char *before,
...
+	l = (int)(str-pstr); /* one step back => return last valid char */
+    }
+    if (num) *num = i - 1;
+    if (len) *len = l;
```
ThomasAdam, line 2237: Style, put the conditionals on their own line, after if()...

ONykyf:
Removed.

ONykyf:
@ThomasAdam Unsafe `strndup()` replaced with `fxstrndup() `analogous to `fxstrdup() `but restricted.

ThomasAdam:
> @ThomasAdam Unsafe `strndup()` replaced with `fxstrndup() `analogous to `fxstrdup() `but truncated if necessary.

No thanks.  I don't think we need this.

> In fact getpwuid.c also contains `strndup() `alongside `fxstrdup()`. Maybe replace it there as well?

getpwuid.c is a compat file found elsewhere, it's not changing to fit in with any wrapper functions.

The point of the `fx*()` wrappers is mostly for portability; `fxmalloc()` is perhaps the only exception to this.

You may as well just remove this latest commit, and make the following adjustment:

```
item->input.init_value = fxmalloc(var_l + 1);
strlcpy(...);
```

I also think we need to look at `item->input_init_value` vs `item->input.value`.

ONykyf:
@ThomasAdam Are there any more adjustments to be made?

PR #1214, opened 2025-06-19
Fvwm: Add configuration option to set FVWM_USERDIR

ONykyf:
Currently $HOME/.fvwm user dir can be changed with FVWM_USERDIR environment variable. It is convenient to change it at compille time, say, to $HOME/.fvwm3, to simplify peaceful coexistence of FVWM2 and FVWM3.

### DIFF
--- a/fvwm/fvwm3.c
+++ b/fvwm/fvwm3.c
@@ -1729,7 +1729,7 @@ int main(int argc, char **argv)
 	{
 		char *s;
 
-		xasprintf(&fvwm_userdir, "%s/.fvwm", home_dir);
+		xasprintf(&fvwm_userdir, "%s/" FVWM_USERDIR, home_dir);
 		/* Put the user directory into the environment so it can be used
 		 * later everywhere. */
 		xasprintf(&s, "FVWM_USERDIR=%s", fvwm_userdir);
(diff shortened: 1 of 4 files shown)


ThomasAdam:
Note as well that:

```
INITIALIZATION
       During initialization, fvwm searches for a configuration file which
       describes key and button bindings, and many other things. The format
       of these files is described later. Fvwm first searches for
       configuration files using the command

           Read config

       This looks for file config in $FVWM_USERDIR and $FVWM_DATADIR
       directories, as described in Read. If this fails more files are
       queried for backward compatibility. Here is the complete list of all
       file locations queried in the default installation (only the first
       found file is used):

           $HOME/.fvwm/config
           /usr/local/share/fvwm/config

           $HOME/.fvwm/.fvwm2rc
           $HOME/.fvwm2rc
           /usr/local/share/fvwm/.fvwm2rc
           /usr/local/share/fvwm/system.fvwm2rc
           /etc/system.fvwm2rc

       Please note, the last 5 locations are not guaranteed to be supported
       in the future.
```

That's from `man fvwm3all`.

We're not about to bugger about with this, there's already means of working around things, including the envvar check of `FVWM_IS_FVWM3` is well, at run-time.

I don't want anything happening to this stuff at compile time.

PR #1216, opened 2025-06-19
FvwmPager: Allow translation of "Desk %d" label via Gettext

ONykyf:
"Desk %d" label at the top of FvwmPager comes always in English, and can be easily localized in *.po (and then *.mo) translation file.

### DIFF
--- a/modules/FvwmPager/x_pager.c
+++ b/modules/FvwmPager/x_pager.c
@@ -31,6 +31,7 @@
 #include <stdbool.h>
 
 #include "libs/FShape.h"
+#include "libs/FGettext.h"
 #include "libs/Module.h"
 #include "libs/FEvent.h"
 #include "fvwm/fvwm.h"


modules/FvwmPager/init_pager.c
```diff
@@ -246,6 +247,9 @@ void initialize_fonts(void)
 	/* Initialize fonts. */
 	FlocaleInit(LC_CTYPE, "", "", "FvwmPager");
 
+	/* Initialize translations. */
+	FGettextInit("fvwm3", LOCALEDIR, "FvwmPager");
+
```
somiaj, line 252: Please don't do this, we do not need to create a .pot file for FvwmPager that contains one string, instead add this to the .pot file for fvwm3. I have moved all the translations into a single file since they are overall quite small in fvwm.

```
FGettextInit("fvwm3", LOCALEDIR, "fvwm3");
```

somiaj:
You should also be more complete on your pull requests, FvwmPager now outputs the 'Monitor RANDRNAME' string if configured, that should also be translated. I can't think of any other strings that might need translated off the top of my head, but you should add those too. Also this pr will need to follow the instructions to update the main fvwm3.pot file with any new strings.

modules/FvwmPager/init_pager.c
ONykyf: Nobody is going to create a separate *.pot file. The name of this file is determined with the first argument `domain`, which is "fvwm3" everywhere. The third argument `module` is just used in log messages, hence all translated messages are collected to fvwm3.pot. All my translations to FvwmForm and FvwmPager are in fvwm3.po, and everything works OK, You can verify it yourself/

I have notived that `module` is also set to "fvwm3" in the existing code, but there is no point for this, we lose additional debug info.

somiaj: Ahh, yea, I guess the second string is never actually used in any meaningful way. When I updated all the translation stuff I did away with separate translations per modules as it was way harder to maintain, and I just changed all the calls to what I suggested. Please just update it to be consistent with how I have it setup for other modules to use this wrapper. Though the actual string you send for the module name has no actual effect.

somiaj:
> Could You point me to the place in the code where this string is output? I looked through the FvwmPager code and have found no more strings to translate. As far as I remember, "Monitor " is not output before the name, I checked it specifically (but I might be wrong).

The function which is drawing the labels is `draw_desk_label` in `x_pager.c`. I guess I was incorrect, the names are just RandR names, so there is nothing to be translated. But You may want to consider that there are also a `small` alternative to the labels that appear with small pagers, for desks, the label will just be `Dn` instead of `Desk n` (or whatever is set). You may want to translate that too?

ONykyf:
> > It is updated because `../modules/FvwmPager/FvwmPager.c` is added to POTFILES and fvwmpo.sh. It also works well in my setup and build script.
> 
> I saw, I was just saying run `./fvwmpo.sh`as part of this PR, so any new translation strings you have added are included in the `.pot`, vs waiting for some translator (which are very rare in fvwm) to come along and do that.

Ok, I will do it now.

modules/FvwmPager/init_pager.c
ONykyf: Then You might consider just removing the third parameter everywhere, because in this case it is meaningless anyway.

somiaj: I agree, but at the time that I combined all the pot files, I just went with what is there. Problem with such an old project is there are lots of things like this, but they are working, so it really isn't that important.

ONykyf: I have searched the code through, and `FGModuleName` variable, which is set by `module` parameter, is used only once, so no real profit, and it can be safely ignored (or removed, which is even better), so let it be "fvwm3".

modules/FvwmPager/init_pager.c
```diff
@@ -60,8 +61,8 @@ void init_fvwm_pager(void)
 
 	/* Run initializations */
 	initialize_mouse_bindings();
-	initialize_desks_and_monitors();
 	initialize_fonts();
+	initialize_desks_and_monitors();
```
ThomasAdam, line 65: Why?

ONykyf: Because `initialize_fonts()` contains localization and (since now) translation, and `initialize_desks_and_monitors()` uses it to set desk label.

modules/FvwmPager/init_pager.c
ThomasAdam: > Then You might consider just removing the third parameter everywhere, because in this case it is meaningless anyway.

Nah, it can stay.  I don't want to change it just for code-churn.

PR #1269, opened 2025-11-29
Add StrMatch test conditional.

somiaj:
Add `Test (StrMatch string1 string2)` conditional that tests if the two strings are equal ignoring case. This is useful to test if a variable expansion, such as $[monitor.current] is equal to a specific string, such as "DP-2", without having to use a shell for this test.


somiaj:
@ThomasAdam I was also considering maybe allowing a pattern match here, such as with EnvMatch, but that is case sensitive, so maybe do a `StrEquals` check and if that fails do a `matchWildcards` check, to allow pattern matching to be used?

PR #1292, opened 2026-02-02
Windowlist: Fix OnlySticky handling

peripherium:
This pull request removes the `WindowList OnlySticky` option.

The `OnlySticky` handling was broken: the flags set by this option overwrote
each other, so the intended behavior never worked. An equivalent and
functional alternative already exists in the form of
`WindowList (sticky)`.

The `OnlySticky` code has been present for more than 20 years without ever
working correctly. It is therefore safe to assume that it has not been used
in practice, and removing it should not break existing user configurations.

This change is purely a cleanup and removes dead and misleading code.

### DIFF
--- a/fvwm/windowlist.c
+++ b/fvwm/windowlist.c
@@ -68,6 +68,7 @@
 #define SHOW_PAGE_Y		(1<<18)
 #define NO_LAYER		(1<<19)
 #define SHOW_SCREEN		(1<<20)
+#define SHOW_ONLY_STICKY     (1<<21)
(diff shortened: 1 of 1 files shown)


ThomasAdam:
> Since the TODO.md list indicates there are too many commands, I decided to remove it. But we can fix it. Perhaps @ThomasAdam has a preference for one of these approaches.

If you write a patch to fix this (it's not broken per se, BTW) I'll probably accept it.

Removing it though?  No...

And although the TODO file mentions deprecating commands, etc., that's not quite the same thing as *options* to existing commands, although even then discretion is advised...

somiaj:
Please squish all your commits together into a single commit.

PR #1294, opened 2026-02-09
FvwmPrompt: use readline directly

### LINKED ISSUE #1270: FvwmPrompt:Restart dies , and doesn't exit gracefully, which hang the terminal FvwmPrompt was started in.
* Fvwm3 version  1.1.4.1.1.3-21-ga15c008f
* gentoo ~amd64  kernel 6.1.154-gentoo

* Linux AMD Ryzen 7 5700U with Radeon Graphics
## Expected Behaviour

I launched a terminal, ran FvwmPrompt and typed Restart + the Enter key

## Actual Behaviour

I expected that after the restart the term will continue to run fine.

The restart went fine. But I got in the term:
>>> INFO[0003] Error reading from UDS: EOF
and the term was hang and didn't reacted to the keyboard anymore.

If I type exit instead of Restart, FvwmPrompt quit gracefully and the  term continue to  work as expected.

I got that both with fvwm-crystal and Xephxr runing a second fvwm3 instance using the default fvwm3 config.

somiaj said: I get the same behavior, the issue isn't that FvwmPrompt quits, it is that FvwmPrompt dies, and doesn't exit gracefully, and since it doesn't exit gracefully the terminal it was running in no longer shows what is being typed into it
and:  typing 'rest' into the terminal got it showing what I typed again, there appears to be a bug there, FvwmPrompt should gracefully quit vs die when its connection to FvwmMFL dies

thomas_adam said: I suspect the issue is that FvwmMFL dies, but FvwmPrompt isn't propagating that.

... and include the output here.

## Extra Information

* Feel free to take a screen capture or video and upload to this issue if you
  feel it would help.

* Attach `$HOME/.fvwm/fvwm3-output.log` from the step above.

ThomasAdam:
ishell isn't being maintained, and we were never getting the good stuff out of it.

Since ishell is written largely on top of readline, we may as well use this directly.

Fixes #1270


somiaj:
My initial tests everything works. FvwmPrompt behaves the same, it now correctly exists when fvwm is reset. Also thanks for changing readline library, now it appears that every golang library is also currently in Debian, so this means I should be able to get FvwmPrompt in the Debian package.

PR #1296, opened 2026-02-18
FvwmConsole: remove in favour of FvwmPrompt

ThomasAdam:
FvwmConsole has always been ear-marked for deletion.

FvwmPrompt has been around long enough that downstream can package this
as they see fit.

### DIFF
--- a/doc/FvwmPrompt.adoc
+++ b/doc/FvwmPrompt.adoc
@@ -29,15 +29,6 @@ Press ^D or type 'exit' to end this session
 >>>
 ....
 
-== Prerequisites
-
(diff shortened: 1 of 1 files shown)


doc/meson.build
```diff
@@ -23,6 +22,10 @@ adoc_sources = [
     'FvwmScript.adoc',
 ]
 
+if golang.found()
+    adoc_sources += 'FvwmPrompt.adoc'
+endif
```
somiaj: It appears that you are no longer removing to the link to FvwmPrompt.html from the index page if golang is not enabled with the changes.

somiaj:
There is something not right with html generation. Every html file that gets generated has the context of `index.html`, except `index.html` which has the contents of `fvwm3styles.html`.

PR #1309, opened 2026-04-25
Fixes #818 icon redraw corruption due to dropped Expose events

### LINKED ISSUE #818: Icons on the root window don't always refresh when damaged or displayed
## Upfront Information

the output.

fvwm3 1.0.7 (1.0.6a-1-g8670b593)
with support for: ReadLine, XPM, PNG, SVG, Shape, XShm, SM, Bidi text,
XRandR, XRender, XCursor, XFT, NLS

Fedora 36 (but also on Fedora 37)

* Platform (run: `uname -sp`)
Linux x86_64

## Expected Behaviour

I use icon-on-root iconified windows with 'IconBox none'. Some of the time when an icon is created or damaged, fvwm3 doesn't repaint the icon and/or the icon title. One of the things that will trigger this is de-iconifying an icon (via a mouse click on it in my setup) where the icon or the expanded icon title overlaps with another icon beside it. The section that the icon or icon title overlapped will not be repainted. Another reproduction appears to be creating iconified windows in a burst. I have a test Wish program for a libX11 related issue that opens two iconified windows in fixed position, and when I run it under fvwm3 the second icon isn't painted either for the icon or for the icon title.

(I use 'Style "*" Icon xlogo32', so xlogo32 is the icon used for both icons. This occurs for other programs with their own icons, such as Firefox.)

Here is that program:

```
#!/usr/bin/wish
wm title . "iconify example"
pack [button .b2 -text "exit" -command exit]
wm iconposition . 500 500
wm iconify .

toplevel .sec
pack [label .sec.l -text "This is a label"]
pack [button .sec.ok -text "OK" -command {puts "OK"}]
wm title .sec "Another icon"
wm iconposition .sec 600 600
wm iconify .sec
```
Bringing the mouse cursor to such a non-repainted icon or title will repaint it properly, as will various things that cause large scale refreshes (switching to another virtual screen, putting a window over the area of the icons, etc).

Unfortunately nothing is reported in fvwm3's logging. This issue didn't happen in fvwm2.

JRRandall:
### What does this PR do?

Fixes an issue where icon windows could fail to repaint when overlapping or adjacent icons changed state (e.g., de-iconifying with long titles).

Expose events were previously processed through a global weed/handle pipeline, which could coalesce or invalidate events across different windows. This caused repaint events for one window to be dropped when another window generated Expose events.

This PR replaces the global handling with per-window processing by iterating over queued Expose events and coalescing them only within the same window using flush_accumulate_expose().

This preserves event coalescing while restoring correct per-window repaint behavior.

## Issue number(s)
```
Fixes #818
```

### DIFF
--- a/fvwm/events.c
+++ b/fvwm/events.c
@@ -268,26 +268,6 @@ static int _pred_weed_accumulate_expose(
 	return 1;
 }
 
-static int _pred_weed_is_expose(
-	Display *display, XEvent *event, XPointer arg)
-{
-	return (event->type == Expose);
-}
-
-static int _pred_weed_handle_expose(
-	Display *display, XEvent *event, XPointer arg)
-{
-	if (event->type == Expose)
-	{
-		dispatch_event(event);
-		return 1;
-	}
-	else
-	{
-		return 0;
-	}
-}
-
 static int _pred_weed_event_type(
 	Display *display, XEvent *event, XPointer arg)
 {
@@ -4547,11 +4527,18 @@ void flush_accumulate_expose(Window w, XEvent *e)
 void handle_all_expose(void)
 {
 	void *saved_event;
+  XEvent e;
 
 	saved_event = fev_save_event();
 	FPending(dpy);
-	FWeedAndHandleIfEvents(dpy, _pred_weed_is_expose,
-			       _pred_weed_handle_expose, NULL);
+  while (FCheckTypedEvent(dpy, Expose, &e))
+  {
+    /* Merge any additional queued Expose events for this same
+     * window into one bounding box and dispatch a single redraw.
+     */
+    flush_accumulate_expose(e.xexpose.window, &e);
+    dispatch_event(&e);
+  }
 	fev_restore_event(saved_event);
 
 	return;


ThomasAdam:
Hi @JRRandall 

Thanks, but I don't think this is the correct thing to do.  We used to use a similar approach, and the reason we rewrote it was because we had the opposite issue of missing Expose events.

I've not had time (and won't have time) to look into this, but it's worthwhile going back tp look at the libx11 changes which ultimately introduced this issue and to work back from there.

But I don't think this change is the correct one.

JRRandall:
@ThomasAdam  Thanks for taking the time to review this, I really appreciate the feedback and the context around the thread-safety changes in libX11.  It sounds like maybe it changed more than just timing...

While continuing to debug this, I tried temporarily bypassing Expose weeding entirely as an experiment, and that consistently resolved the repaint issue. That seems to suggest the problem is tied specifically to how Expose events are being coalesced rather than the broader event handling.

I’m going to take another pass at this with that in mind and see if I can come up with a more targeted fix that preserves the intended behavior.

PR #1310, opened 2026-04-26
events: restrict Expose weeding scope to prevent cross-window coalescing

### LINKED ISSUE #818: Icons on the root window don't always refresh when damaged or displayed

JRRandall:
### What does this PR do?
Expose events are currently processed within a single weeding pass, which can coalesce events across multiple windows.  In some cases, this can result in repaint obligations being dropped when Expose activity from one window affects another.

This change keeps the existing weeding pipeline intact, but returns both match (`rc & 1`) and stop (`rc & 2`) for Expose events.  This limits each weeding pass to a smaller scope so Expose events are not coalesced across unrelated windows.

This preserves existing event handling behavior while preventing cross-window interaction during Expose coalescing.

### Screenshots (if applicable)
N/A (behavioral fix; verified via `xev -id` and icon repaint testing)

### Issue number(s)
Fixes #818

### DIFF
--- a/fvwm/events.c
+++ b/fvwm/events.c
@@ -271,7 +271,14 @@ static int _pred_weed_accumulate_expose(
 static int _pred_weed_is_expose(
 	Display *display, XEvent *event, XPointer arg)
 {
-	return (event->type == Expose);
+  if (event->type == Expose)
+  {
+    /* Include Expose in weeding (rc & 1) and stop scanning early (rc & 2)
+     * to avoid coalescing Expose events across different windows.
+     */
+    return (1 /* match */ | 2 /* stop */);
+  }
+  return 0;
 }
 
 static int _pred_weed_handle_expose(


JRRandall:
@ThomasAdam This patch is intentionally minimal and scoped to the observed regression: it restricts Expose coalescing to a single window per weeding pass without changing the overall event handling architecture or removing the existing pipeline behavior.  The goal here is strictly to prevent cross-window interaction during Expose processing, which is where the repaint issue was consistently observed.

I’m happy to iterate further if there’s a preferred long-term direction for more structured per-window coalescing, but from the testing so far this resolves the practical regression without introducing broader behavioral changes.

JRRandall:
Hi @siebenmann @ThomasAdam even though we are convinced the upstream changes of XOrg for multithread introduced this issue, I believe I have a proper fix for the FVWM3 behavior in this PR if you've got a moment to review it, or if you can assign it to someone who does.  Thanks!

ThomasAdam:
> Hi @siebenmann @ThomasAdam even though we are convinced the upstream changes of XOrg for multithread introduced this issue, I believe I have a proper fix for the FVWM3 behavior in this PR if you've got a moment to review it, or if you can assign it to someone who does. Thanks!

I think I've said all I need to already.

I don't want to merge this.

I'm sick of fvwm being a "mop up" application for fixes which otherwise live elsewhere.

If you can't lobby the Xorg folks to fix this, maybe the crazy maintainer of Xlibre might be willing instead?

JRRandall:
I understand your concern about carrying upstream workarounds.  My intent wasn't to shift responsibility to FVWM, only to restore correct behavior for users affected by the Xorg/libX11 threading changes until an upstream fix exists.  If you'd still prefer not to merge it, I understand.  I'll keep the patch available for FVWM users and reach out to Enrico about whether it can be addressed in XLibre.

JRRandall:
Trying to do the right thing here https://github.com/orgs/X11Libre/discussions/551

PR #1317, opened 2026-05-10
Fix moving windows from one desktop to another.

tzimisces:
* **What does this PR do?**

It turns out that windows disappear when being moved from one desktop to another, when the current desktop is the desktop the window is to be moved to, and, yes, the original desktop is not the current one ;)

I believe this fixes that problem.

### DIFF
--- a/fvwm/move_resize.c
+++ b/fvwm/move_resize.c
@@ -2246,7 +2246,11 @@ static void _move_window(F_CMD_ARGS, Bool do_animate, int mode)
 		final.x = r.x;
 		final.y = r.y;
 
-		fw->Desk = m->virtual_scr.CurrentDesk;
+		if (fw->Desk != m->virtual_scr.CurrentDesk)
+		{
+			fw->UpdateDesk = m->virtual_scr.CurrentDesk;
+		}
+
 		fw->m = m;
 	}
 	else


fvwm/move_resize.c
```diff
@@ -2246,7 +2246,11 @@ static void _move_window(F_CMD_ARGS, Bool do_animate, int mode)
 		final.x = r.x;
 		final.y = r.y;
 
-		fw->Desk = m->virtual_scr.CurrentDesk;
+		if (fw->Desk != m->virtual_scr.CurrentDesk)
+		{
+			fw->UpdateDesk = m->virtual_scr.CurrentDesk;
+		}
+
```
ThomasAdam, line 2253: No -- I'm not sure this is correct; the current desk gets updated when `fvwm_update_screen()` is called.

ThomasAdam (changes requested):

fvwm/move_resize.c
tzimisces: You mean fvwm_update_monitor? True, but if UpdateDesk is not set correctly, the window will not be made visible? 
In a sense, it is also suggested by the documentation of UpdateDesk that one should use that field in case of a desk-change?

PR #1318, opened 2026-05-10
Fix situation in which edge commands are not triggered.

tzimisces:
* **What does this PR do?**

Situation is as follows. Application window is maximised. One opens a menu in that application. Menu is closed. One moves the pointer to an edge. Edge command is not triggered.

This commit fixes that. Rationale: when refocusing too early, the edge is not considered anymore and not command is triggered.

### DIFF
--- a/fvwm/events.c
+++ b/fvwm/events.c
@@ -2261,13 +2261,6 @@ void HandleEnterNotify(const evh_args_t *ea)
 		char *edge_command = NULL;
 		struct monitor	*m = monitor_get_current();
 
-		if (
-			Scr.UnknownWinFocused != None &&
-			(sf = get_focus_window()) != NULL &&
-			sf == Scr.StolenFocusFvwmWin)
-		{
-			_refocus_stolen_focus_win(ea);
-		}
 		/* check for edge commands */
 		if (ewp->window == m->PanFrameTop.win)
 		{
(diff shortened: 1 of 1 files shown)


ThomasAdam (changes requested):
Hi @tzimisces 

If you do this, you'll cause problems with other windows.

Is this more of a timing issue, can you confirm?

fvwm/events.c
```diff
@@ -2259,15 +2259,8 @@ void HandleEnterNotify(const evh_args_t *ea)
 	if (is_pan_frame(ewp->window))
 	{
 		char *edge_command = NULL;
-		struct monitor	*m = monitor_get_current();
+		struct monitor  *m = monitor_get_current();
```
ThomasAdam: What's this change for?

fvwm/events.c
```diff
@@ -4121,6 +4123,7 @@ void dispatch_event(XEvent *e)
 
 	switch (e->type - randr_event) {
 	case RRScreenChangeNotify:
+		fvwm_debug(__func__, "RRScreenChangeNotify (%d).", e->type);
```
ThomasAdam: Remove this.

ThomasAdam (changes requested):

fvwm/events.c
tzimisces: Beats me ;) I'll revert that.

tzimisces:
No, I don't think it's a timing issue, and it's deterministic.

What I observe for GTK-applications: 
* the application is 100% maximised,
* one opens and closes a dropdown menu,
* one moves the pointer so it accesses the pan-frame,
* no edge-command is triggered,

The above sequence results in the pointer not leaving the application window. Then, for some reason, Fvwm's state suggests that something should be done about a window whose focus is stolen.
That's fine with me, but in this case, it will result in the wrong window being evaluated for the presence of an edge command, which is a pity, because I use that quite extensively :)

For as far as I can see, the only consequence of this change is the fact that the right "window" is evaluated for the presence of an edge command, and that after that, things proceed like before. Thing is, the "ewp"-pointer changes after the refocus, so the pan frame is not considered anymore.

What problems would this change yield for other windows?
Is there a relation with HandleFocusOut?

ThomasAdam:
Hi @tzimisces 

Can you squash these two commits into one, please?

tzimisces:
No problem!

ThomasAdam:
Thanks.

